PR #1, opened 2017-11-15
Add ansible setup playbook on openstack infra

dincho:
- [x] Re-encrypt travis private SSH key
- [x] OpenStack ENV configuration


deploy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -ev
+
+# Free Travis version does not support SSH key pairs, thus using encrypted files for SSH deploy key
+# https://docs.travis-ci.com/user/encrypting-files/
+openssl aes-256-cbc -K $encrypted_989f4ea822a6_key -iv $encrypted_989f4ea822a6_iv -in ansible/files/travis/master_key.enc -out master_key -d
```
lucafavatella: This line says `-out master_key`. Lines after say `/tmp/deploy_rsa` (see lines 8 and 9).
It looks like names shall be the same? (It looks copy-paste from https://github.com/aeternity/epoch/pull/362/files?w=1#diff-35ea51baf1fe7f0142ad5f950855dde0R83)

And also, how does this related to `ansible/files/travis/master_key.pub`?

ansible/setup.yml
```diff
@@ -0,0 +1,60 @@
...
+  vars:
+    project_user: epoch
+    devops:
+      - dincho
+    developers:
+      - sennui
+      - velzevur
+
+  tasks:
+    - name: Create epoch daemon user
```
lucafavatella: I guess here you wanted to use `{{ project_user }}`. No need to change PR of course.

deploy.sh
```diff
@@ -0,0 +1,13 @@
...
+chmod 600 /tmp/deploy_rsa
+eval "$(ssh-agent -s)"
+ssh-add /tmp/deploy_rsa
+
+# Setup all epoch nodes with Ansible
+cd ansible && ansible-playbook --limit=epoch setup.yml
```
lucafavatella: I see this implements what is documented in [README.md](https://github.com/aeternity/infrastructure/pull/1/files#diff-04c6e90faac2675aa89e2176d2eec7d8R21)

But I see the other preceding instructions, documented in README.md, neither in the `.travis.yml` nor in this `deploy.sh`. Why?
Those instructions are:
* `pip install -r ansible/pip-requirements.txt` - lack of this instruction would mean depending on packages already installed on Travis hence potentially changing over time
* `source ~/my/secrets/openstack.rc` - I understand this is preparatory for the next one;
* `ansible-inventory -i inventory/openstack.yml --list` - this would be useful for logging in CI hosts being recognized.

ansible/setup.yml
```diff
@@ -0,0 +1,60 @@
...
+        - nodes-ssh-keys
+
+    - name: Add deploy@travis SSH key to authorized keys
+      authorized_key:
+        user: "{{ project_user }}"
+        key: "{{ lookup('file', 'files/travis/deploy_key.pub') }}"
+      tags:
+        - deploy-ssh-keys
+
+    - name: Install dev tools
```
lucafavatella: I see no `apt update && apt upgrade`. Is that subject to separate periodic procedure?

README.md
```diff
@@ -1 +1,22 @@
-# infrastructure
\ No newline at end of file
+# Infrastructure management automation for æternity nodes
+
+## Ansible Deploy
+
+This implementation supports OpenStack dynamic inventory by using [Ansible 2.4 OpenStack inventory plugin](https://docs.ansible.com/ansible/devel/plugins/inventory/openstack.html).
+To install it's dependencies with `pip`:
+```bash
+pip install -r ansible/pip-requirements.txt
```
lucafavatella: pip-requirements.txt or requirements.txt ?

deploy.sh
lucafavatella: I see that the documented `~/my/secrets/openstack.rc` may include pointers to the environment so you may want not to put it in clear in the repo and rather store it as encrypted with Travis encrypted var? (at the moment it is not in the repo at all.)

For the same reason you may want not to insert `ansible-inventory -i inventory/openstack.yml --list` logging call?

ansible/inventory/openstack.yml
```diff
@@ -0,0 +1,3 @@
+plugin: openstack
+inventory_hostname: 'name'
```
lucafavatella: What does this `inventory_hostname: 'name'` mean?

It is unclear which hosts are pointed to by these scripts, and who creates those hosts. What I understand is these scripts assume the hosts are already created, and idempotently configure them (ensure keys are configured, ensure git is installed, ...).

.travis.yml
```diff
@@ -0,0 +1,13 @@
+dist: trusty
+language: python
+script:
+  - ansible-lint ansible/setup.yml
+deploy:
+  - provider: script
+    script: deploy.sh
```
lucafavatella: If CI kicks in (e.g. when this PR is merged - if Travis is enabled on this repo) I expect this job to fail.

Please add:
* One line comment on what this script is meant to do;
* How this script integrate in the Travis CI flow. It looks like that if master of this repo is changed e.g. you add a developer to the list, then the CI is meant to kick in and update the configuration of the (already existing) hosts.

ansible/setup.yml
dincho: Actually no, I meant the user for the daemon bin/epoch. Probably my english wording is not good enough?

dincho: Actually no, I meant the user for the daemon bin/epoch. Probably my english wording is not good enough?

ansible/setup.yml
```diff
@@ -0,0 +1,60 @@
...
+    - name: Create epoch daemon user
+      user:
+        name: "{{ project_user }}"
+        shell: /bin/bash
+      tags:
+        - project-user
+
+    - name: "Setup root SSH authorized keys"
+      authorized_key:
+        user: "{{ ansible_user }}"
```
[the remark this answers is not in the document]
dincho: It's the user the ansible uses to connect to the host, in practice it's the same as remote_user. I cannot currently find a concrete documentation on this to reference it.

deploy.sh
dincho: README is meant more for manually testing, I'll update it.

- requirements.txt is [auto-picked by travis](https://docs.travis-ci.com/user/languages/python/#pip)
- indeed ~/my/secrets/openstack.rc is for local testing and it contains the sensitive openstack configuration. This configuration is set in travis environment settings (marked as secure)
- this information is still dumped during ansible playbook execution, so no need for additional requests here

ansible/inventory/openstack.yml
dincho: As in the README, this playbook uses OpenStack dynamic inventory and limit the execution based on configured host groups.

Indeed you got it right, this playbook just configure an already created groups of hosts.

inventory_hostname just make it use instance name instead of instance id (which is ugly UUID). There is a [link to this docs already](https://docs.ansible.com/ansible/devel/plugins/inventory/openstack.html) in the README but I'll add it here too.

deploy.sh
dincho: fixed in 0f71228817b545dd527531951b40a74d1373dd06

I'm tracking master_key.pub just for the sake of completeness. It's not used yet in the automation.

ansible/setup.yml
dincho: I added automatic apt-get update in 712633aed4166602c389ecdce0de374cff189827

However I prefer to keep upgrades as periodic procedures.

README.md
dincho: fixed in efa9504c7fd927de7c20664571ab09b6fe037207

ansible/inventory/openstack.yml
dincho: I added more info in efa9504c7fd927de7c20664571ab09b6fe037207

.travis.yml
dincho: I added more info in README in efa9504c7fd927de7c20664571ab09b6fe037207. I think it's better place to write about workflow instead in travis config.

dincho:
> Please amend the usage of authorized_key in the Ansible script as current usage does not support removing keys (e.g.: a developer decides to change his/her github key because fears compromised - these scripts do not remove that; e.g. a developer moves away from the project - simply removing developer from ansible variable developers does not remove those keys).
> 
> It looks like scripts need to be amended by usage of option exclusive. This means all keys would need to be concatenated in one file.
> 
> In alternative to the above, please at least mark need for such update with a clear comment for each usage of authorized_key.

I added comments in 06a537d. I'll investigate what's the options to handle this cases because "exclusive" looks a little bit hairy.

ansible/setup.yml
lucafavatella: No, it is ok. Thanks for explanation.

PR #7, opened 2017-11-27
Datadog integration

dincho:
- install datadog agent to all epoch nodes
- add ansible callback plugin to feed datadog with ansible stats

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,35 @@ To install dependencies use `pip`:
 pip install -r requirements.txt
 ```
 
+You also need to install the Ansible roles used in playbooks:
+```bash
+ansible-galaxy install -r ansible/requirements.yml
+```
+
 ## Credentials setup
 
+### OpenStack
 You should make sure [OpenStack credentials are set](https://docs.openstack.org/python-openstackclient/latest/configuration/index.html#environment-variables)
 either by environment variables or clouds.yml file.
 
 ```bash
 source ~/my/secrets/openstack.rc
+```
+
+### DataDog
+Ansible is feeding datadog via callback plugin on each run.
+You should create a callback configuration with your DataDog API key before running any playbook.
+
+```bash
+echo "api_key: my_datadog_api_key_woof" > ansible/callback_plugins/datadog_callback.yml
+```
+
+Also monitoring.yml requires `datadog_api_key` variable to be set.
+You can use any supported ansible method, for example you can use local vars (not tracked in .gitignore:
+
+```bash
+mkdir -p ansible/group_vars/all
+echo "datadog_api_key: my_datadog_api_key_woof" >> ansible/group_vars/all/vars.local.yml
 ```
 
 ## Infrastructure orchestration
@@ -46,4 +68,5 @@ Setup environments infrastructure by running:
 cd ansible
 ansible-playbook environments.yml
 ansible-playbook setup.yml
+ansible-playbook monitoring.yml
 ```
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ echo "api_key: my_datadog_api_key_woof" > ansible/callback_plugins/datadog_callb
 ```
 
 Also monitoring.yml requires `datadog_api_key` variable to be set.
-You can use any supported ansible method, for example you can use local vars (not tracked in .gitignore:
+You can use any supported ansible method, for example you can use local vars (not tracked in .gitignore):
 
 ```bash
 mkdir -p ansible/group_vars/all
--- a/ansible/callback_plugins/datadog_callback.py
+++ b/ansible/callback_plugins/datadog_callback.py
@@ -1,0 +1,280 @@
+# Source: https://github.com/DataDog/ansible-datadog-callback
+# Tag: 1.0.1
+
+import getpass
+import os.path
+import time
+
+import datadog
+import yaml
+
+try:
+    # Ansible v2
+    from ansible.plugins.callback import CallbackBase
+    from __main__ import cli
+except ImportError:
+    # Ansible v1
+    CallbackBase = object
+    cli = None
+
+
+class CallbackModule(CallbackBase):
+    def __init__(self):
+        # Read config and set up API client
+        api_key, url = self._load_conf(os.path.join(os.path.dirname(__file__), "datadog_callback.yml"))
+        datadog.initialize(api_key=api_key, api_host=url)
+
+        self._playbook_name = None
+        self._start_time = time.time()
+        self._options = None
+        if cli:
+            self._options = cli.options
+
+        # self.playbook is either set by Ansible (v1), or by us in the `playbook_start` callback method (v2)
+        self.playbook = None
+        # self.play is either set by Ansible (v1), or by us in the `playbook_on_play_start` callback method (v2)
+        self.play = None
+
+    # Load parameters from conf file
+    def _load_conf(self, file_path):
+        conf_dict = {}
+        with open(file_path, 'r') as conf_file:
+            conf_dict = yaml.load(conf_file)
+
+        return conf_dict.get('api_key', ''), conf_dict.get('url', 'https://app.datadoghq.com')
+
+    # Send event to Datadog
+    def _send_event(self, title, alert_type=None, text=None, tags=None, host=None, event_type=None, event_object=None):
+        if tags is None:
+            tags = []
+        tags.extend(self.default_tags)
+        priority = 'normal' if alert_type == 'error' else 'low'
+        try:
+            datadog.api.Event.create(
+                title=title,
+                text=text,
+                alert_type=alert_type,
+                priority=priority,
+                tags=tags,
+                host=host,
+                source_type_name='ansible',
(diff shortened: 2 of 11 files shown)


ansible/callback_plugins/datadog_callback.py
```diff
@@ -0,0 +1,280 @@
+# Source: https://github.com/DataDog/ansible-datadog-callback
```
dincho: I don't like to include this in the repo, but I don't see a better method to source it

ansible/monitoring.yml
```diff
@@ -0,0 +1,37 @@
...
+          # The Agent just needs one item in instances in order to enable the check.
+          # The content of the item doesn’t matter.
+          - foo: bar
+      process:
+        init_config:
+        instances:
+          - name: ssh
+            search_string: ['sshd']
+          - name: epoch
+            search_string: ['beam.smp']
```
dincho: I'm not sure if this is the best process name to monitor

deploy.sh
```diff
@@ -9,7 +9,12 @@ chmod 600 /tmp/master_rsa
...
 
+# Unfortunately the callback does not support environment vars but only config
+echo "api_key: ${datadog_api_key:?}" > ansible/callback_plugins/datadog_callback.yml
+
 # Setup environments and all epoch nodes with Ansible
 cd ansible
+ansible-galaxy install -r requirements.yml
 ansible-playbook environments.yml
 ansible-playbook --limit=epoch setup.yml
+ansible-playbook --limit=epoch --extra-vars "datadog_api_key=${datadog_api_key:?}" monitoring.yml
```
dincho: limit should be changed to `integration:uat:&epoch`

dincho: changed in c7c5bc5

README.md
```diff
@@ -14,15 +14,37 @@ To install dependencies use `pip`:
...
+### DataDog
+Ansible is feeding datadog via callback plugin on each run.
+You should create a callback configuration with your DataDog API key before running any playbook.
+
+```bash
+echo "api_key: my_datadog_api_key_woof" > ansible/callback_plugins/datadog_callback.yml
+```
+
+Also monitoring.yml requires `datadog_api_key` variable to be set.
+You can use any supported ansible method, for example you can use local vars (not tracked in .gitignore:
```
lucafavatella: Missing closing parenthesis (optional for this PR).

ansible/monitoring.yml
lucafavatella: This looks correct. Though I see two potential issues:
1. If operator needs to attach to node for inspecting state, I expect he/she will use e.g. `bin/epoch remote_console` that rely on `epmd` process not having died since the epoch node started;
2. If operator needs to attach to node for inspecting state e.g. `bin/epoch remote_console` this will start a second beam.smp process: does datadog get confused?

I am checking solutions for the two problems.

ansible/monitoring.yml
```diff
@@ -0,0 +1,42 @@
...
+      - "env:{{ datadog_env }}"
+      - "lsb:{{ ansible_lsb.description }}"
+      - "flavor:c{{ ansible_processor_vcpus }}.m{{ (ansible_memtotal_mb/1024)|round|int }}"
+    datadog_config:
+      apm_enabled: false
+      use_dogstatsd: yes
+      process_agent_enabled: true
+      hostname: "{{ ansible_fqdn }}"
+      tags: "{{ datadog_tags|join(',') }}"
+    datadog_checks:
```
lucafavatella: @dincho Where is the documentation of the reference of this `datadog_checks`? (I see multiple versions of the agent, and did not find clear pointer to documentation.)

ansible/callback_plugins/datadog_callback.py
lucafavatella: `curl -LO ...`  of the raw versioned link on github? Though I am now sure that would be an improvement.

Could you please amend the link to be a versioned link - so that the exact version of the script is clear.

ansible/monitoring.yml
lucafavatella: > 2. If operator needs to attach to node for inspecting state e.g. bin/epoch remote_console this will start a second beam.smp process: does datadog get confused?

Below is an example of the two beam.smp processes as returned by `ps aux` - the best difference appears to be the `-boot /tmp/node/releases/0.3.3/start` part (after beam.smp):
```
ubuntu    7089  0.4  7.6 2695292 77656 pts/1   Ssl+ 10:58   0:07 /tmp/node/erts-9.0.1/bin/beam.smp -K true -A30 -SDio 10 -- -root /tmp/node -progname tmp/node/bin/epoch -- -home /home/ubuntu -- -boot /tmp/node/releases/0.3.3/start -mode embedded -boot_var ERTS_LIB_DIR /tmp/node/erts-9.0.1/../lib -config /tmp/node/releases/0.3.3/sys.config -sname epoch -setcookie epoch_cookie -- console
ubuntu    8176  0.1  2.7 2629512 28440 pts/0   Sl+  11:24   0:00 /tmp/node/erts-9.0.1/bin/beam.smp -- -root /tmp/node -progname erl -- -home /home/ubuntu -- -sname remsh42e4e36b-epoch@ubuntu-xenial -remsh epoch@ubuntu-xenial -boot start_clean -boot_var ERTS_LIB_DIR /tmp/node/erts-9.0.1/../lib -setcookie epoch_cookie -hidden -kernel net_ticktime 60
```

I did not find the reference of these `datadog_checks` so not sure how to express this in the config.

lucafavatella: > 1. If operator needs to attach to node for inspecting state, I expect he/she will use e.g. bin/epoch remote_console that rely on epmd process not having died since the epoch node started;

This looks like not a monitoring/datadog issue but rather something that shall be done in the init supervision tree. I suggest amending datadog config with monitoring if epmd process is up, and warning (not error) if that is not the case.

ansible/callback_plugins/datadog_callback.py
dincho: Well, while downloading the source code will not make a difference in the travis workflow, it will add additional setup step for devops that use the repo manually (and during testing).

I prefer to keep the workflow as simple as `git clone ThisRepo; ansible-playbook ...`

Other solution would be to add https://github.com/DataDog/ as submodule and symlink the callback script in the ansible dir, still I don't see much improvement.

ansible/monitoring.yml
dincho: I assume that more than one epmd process is not expected?

Also why warning instead of error ?

ansible/monitoring.yml
dincho: I fixed the DD role version to [1.4.0](https://github.com/aeternity/infrastructure/pull/7/files#diff-74a84d68b99baca3e0a906aba92bd4c9R2). So the docs should be [this](https://github.com/DataDog/ansible-datadog/blob/1.4.0/README.md), however this role just spit in a conf file that you provide as dictionary.

Agents configuration references/docs can be found [here](https://docs.datadoghq.com/integrations/)

dincho:
@lucafavatella Last time that I tried to use more specific search_string in the checks failed for some reason, so I'm not sure we can check for `-boot /tmp/node/releases/0.3.3/start` without further investigation.

I'm switching the check to `epmd`

ansible/monitoring.yml
lucafavatella: > I assume that more than one epmd process is not expected?

Correct.

lucafavatella: > Also why warning instead of error? Missing epmd will render the node in "non-running" state.

I understand that if epmd died the node will keep running and working. Am I missing something? The issue will be for operator not being able to use `bin/epoch ping` or `bin/epoch remote_console`.

lucafavatella:
> I'm switching the check to epmd

Please keep the check on beam.smp though (just ensuring we are on same page).

ansible/callback_plugins/datadog_callback.py
dincho: Source URL changed in feb25216fafeace0dc5e1647ed4a8f28a5aedbf3

README.md
dincho: added in feb25216fafeace0dc5e1647ed4a8f28a5aedbf3

dincho:
> I'm switching the check to epmd
> Please keep the check on beam.smp though (just ensuring we are on same page).

Well, looks like we wasn't  :) c66842fc8d33f957fc41fc10b86e6579e886e6e8. Reverting.

dincho:
> I'm switching the check to epmd
> Please keep the check on beam.smp though (just ensuring we are on same page).
> Well, looks like we wasn't :) c66842f. Reverting.

See 76900af

ansible/monitoring.yml
dincho: I don't have much understanding how erlang handle it's processes, so I assume your comment is correct.

So, you basically asked me to add additional check for empd (warning) next to beam.smp ?

dincho: see 183187f

PR #9, opened 2017-11-28
Fix openstack environments

dincho:
The "name" parameter looks like reserved or it's just confusing python calls, so I renamed it to "env_name".

[This is the error](https://travis-ci.org/aeternity/infrastructure/builds/308329315#L659).

Also added dry runs to ensure this is auto checked during travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 dist: trusty
 language: python
 script:
-  - ansible-lint ansible/environments.yml
+  - openstack stack create test -e openstack/test/create.yml -t openstack/ae-environment.yml --enable-rollback --wait --dry-run
+  - ansible-playbook --check -i localhost, ansible/environments.yml
   - ansible-lint ansible/setup.yml
 deploy:
   - provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 dist: trusty
 language: python
+before_script:
+  - ansible --version
+  - openstack --version
 script:
   - openstack stack create test -e openstack/test/create.yml -t openstack/ae-environment.yml --enable-rollback --wait --dry-run
   - ansible-playbook --check -i localhost, ansible/environments.yml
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,7 +1,8 @@
 dist: trusty
 language: python
 script:
-  - ansible-lint ansible/environments.yml
+  - openstack stack create test -e openstack/test/create.yml -t openstack/ae-environment.yml --enable-rollback --wait --dry-run
```
lucafavatella: Missing file openstack/test/create.yml ?

dincho: yep here it is 8e513a0, thanks

.travis.yml
```diff
@@ -1,5 +1,7 @@
 dist: trusty
 language: python
+isntall:
```
lucafavatella: Typo?

dincho: I remove back this build phase (using the default).

.travis.yml
```diff
@@ -1,13 +1,11 @@
...
 language: python
-isntall:
-  - pip install --user -r requirements.txt
 before_script:
   - ansible --version
   - openstack --version
 script:
   - openstack stack create test -e openstack/test/create.yml -t openstack/ae-environment.yml --enable-rollback --wait --dry-run
-  - ansible-playbook --check -i localhost, ansible/environments.yml
+  - ansible-playbook --check -i localhost, -e 'ansible_python_interpreter="/usr/bin/env python"' ansible/environments.yml
```
lucafavatella: Was it defaulting to Python3? What was the symptom?

dincho: It looks like Ansible has [hardcoded python path](https://github.com/ansible/ansible/issues/6345#issuecomment-37107657) to system python, but travis is using virtualenv and [won't find the installed modules](https://travis-ci.org/aeternity/infrastructure/builds/308395778#L990)

PR #11, opened 2017-11-29
Allow 1 node environments

dincho:


dincho:
> I see , for this

Indeed

PR #12, opened 2017-12-05
Integrate Datadog logs agent

dincho:
- update Datadog agent to v6 (beta)
- change tags dict (because of v6)
- add dd-agent to epoch (node) group

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ script:
 deploy:
   - provider: script
     script: ./deploy.sh
+    skip_cleanup: true
     on:
       repo: aeternity/infrastructure
       branch: master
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,6 +13,7 @@ script:
 deploy:
   - provider: script
     script: ./deploy.sh
+    skip_cleanup: true
```
lucafavatella: Why?

dincho: This is additional fix to https://travis-ci.org/aeternity/infrastructure/builds/311803848#L1087
We setup the callback configuration in [script phase](https://github.com/aeternity/infrastructure/blob/630f2f1b1af1000f19250a74f274752c9c9caedc/.travis.yml#L8) however travis does the git reset before deploy if we don't disable that with skip_cleanup.

lucafavatella: Thanks for explanation. I see I do not need to re-approve.

PR #13, opened 2017-12-05
Increase UAT nodes to 13

dincho:


ansible/setup.yml
```diff
@@ -35,7 +35,8 @@
         key: https://github.com/{{ item }}.keys
       with_items: "{{ devops }}"
       tags:
-        - master-ssh-keys
```
lucafavatella: These changed tags seem not used so ok to amend them.

ansible/environments.yml
```diff
@@ -48,7 +48,7 @@
         parameters:
           env_name: uat
           meta_groups: "epoch,uat"
-          node_count: 3
+          node_count: 13
```
lucafavatella: My only concern is whether all the (other) nodes i.e. 12 nodes end up in peers in the config https://github.com/aeternity/epoch/blob/e44529295d2b3f323795149dd979af69f28a8784/deployment/ansible/templates/epoch.yaml.j2#L4

lucafavatella: though that shall be supported (no max in https://github.com/aeternity/epoch/blob/e44529295d2b3f323795149dd979af69f28a8784/apps/aeutils/priv/epoch_config_schema.json#L9) so ok - cc @uwiger

PR #14, opened 2017-12-14
Start, stop and restart nodes

StanGenchev:
Start, stop and restart nodes in a deployment environment by using the manage-node.yml and the commands:
ansible-playbook -i hosts manage-node.yml --tags="start"
ansible-playbook -i hosts manage-node.yml --tags="stop"
ansible-playbook -i hosts manage-node.yml --tags="restart"


ansible/manage-node.yml
```diff
@@ -3,12 +3,19 @@
   vars:
     epoch_exec: "{{ ansible_env.HOME }}/node/bin"
   tasks:
-  - name: start
+  - name: Checking the epoch daemon state
+    shell: "{{ epoch_exec }}/epoch pid | cat"
```
dincho: I prefer "epoch ping", looks more reliable to me

StanGenchev: Yes, I`m using epoch ping in the new one.
I`m currently working on other improvements as well.

PR #15, opened 2017-12-28
Manage Node playbook update

StanGenchev:
Update to the manage node playbook where the tags are replaced with a varieble.
The reason for replacing the tags is error handling.
If the user forgets to specify a tag, ansible may throw an error. With a varieble which has a default value that won`t happen and if the user enters the wrong value when executing the playbook there will be a message saying the input is incorrect.
This playbook is started using the command:
ansible-playbook -i hosts manage-node.yml --extra-vars="c=<action>"
<action> can be 'start', 'stop' or 'restart'.


StanGenchev:
Added reset test net playbook and fixed typos in the manage node playbook.

ansible/manage-node.yml
```diff
@@ -0,0 +1,35 @@
...
+  hosts: all
+  remote_user: epoch
+  vars:
+    project_root: "{{ ansible_env.HOME }}/node"
+    epoch_action: "{{ c | default('none') }}"
+  tasks:
+  - name: Checking if the input in correct
+    debug:
+      msg: "Incorrect input! You must input either 'start', 'stop' or 'restart'"
+    when: epoch_action != "start" and epoch_action != "stop" and epoch_action != "restart"
```
dincho: this task always report as changed, `changed_when: false` should help

ansible/manage-node.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Manage epoch nodes
+  hosts: all
+  remote_user: epoch
+  vars:
+    project_root: "{{ ansible_env.HOME }}/node"
+    epoch_action: "{{ c | default('none') }}"
```
dincho: I really don't like one char variables, let's change it to at least `cmd`

ansible/manage-node.yml
```diff
@@ -0,0 +1,35 @@
...
+  tasks:
+  - name: Checking if the input in correct
+    debug:
+      msg: "Incorrect input! You must input either 'start', 'stop' or 'restart'"
+    when: epoch_action != "start" and epoch_action != "stop" and epoch_action != "restart"
+
+  - name: Check if epoch binary exists
+    stat:
+      path: "{{ project_root }}/bin/epoch"
```
dincho: this task always report as changed

ansible/manage-node.yml
dincho: we also need `ping` cmd

ansible/manage-node.yml
```diff
@@ -0,0 +1,35 @@
...
+    command: "{{ project_root }}/bin/epoch start"
+    when: epoch.stat.exists == True and not ping.stdout == "pong" and epoch_action == "start"
+
+  - name: Stopping the epoch deamon
+    command: "{{ project_root }}/bin/epoch stop"
+    when: epoch.stat.exists == True and ping.stdout == "pong" and epoch_action == "stop"
+
+  - name: Restarting the epoch deamon
+    command: "{{ project_root }}/bin/epoch restart"
+    when: epoch.stat.exists == True and ping.stdout == "pong" and epoch_action == "restart"
```
dincho: Restart command implies starting the node even if it's stopped, which this task unfortunately won't do.

ansible/manage-node.yml
dincho: actually we must fail if the binary does not exists, no?

ansible/reset-net.yml
```diff
@@ -0,0 +1,63 @@
+---
+- name: Reset epoch test-net
+  hosts: all
+  remote_user: epoch
+  vars:
+    project_root: "{{ ansible_env.HOME }}/node"
+  tasks:
+    - name: Check if epoch binary exists
```
dincho: just fail if the binary does not exists

README.md
```diff
@@ -70,3 +70,20 @@ ansible-playbook environments.yml
...
 ansible-playbook monitoring.yml
 ```
+## Manage nodes
+
+Start, stop and restart nodes by running:
+```bash
+cd ansible
+ansible-playbook manage-node.yml --extra-vars="c=start"
+ansible-playbook manage-node.yml --extra-vars="c=stop"
+ansible-playbook manage-node.yml --extra-vars="c=reset"
```
dincho: don't forget to add `ping`

README.md
```diff
@@ -70,3 +70,20 @@ ansible-playbook environments.yml
...
+```
+## Reset test-net nodes
```
dincho: just `## Reset nodes`, its not meant to be run on test-net only and the term is already confusing

dincho:
Add appropriate linting commands in travis config.

dincho:
closing in favour of #17

PR #17, opened 2018-01-18
Manage playbooks

dincho:
- manage-node playbook to ping, start, stop, restart nodes
- reset-net playbook to reset already running [test]net by deleting it's blockchain data and logs

https://www.pivotaltracker.com/story/show/152628820

### DIFF
--- a/ansible/manage-node.yml
+++ b/ansible/manage-node.yml
@@ -4,7 +4,7 @@
   remote_user: epoch
 
   vars:
-    epoch_bin: "{{ ansible_env.HOME }}/node//bin/epoch"
+    epoch_bin: "{{ ansible_env.HOME }}/node/bin/epoch"
     epoch_action: "{{ cmd | default('ping') }}"
     epoch_running: false
 
@@ -62,3 +62,8 @@
       when:
         - epoch_running
         - epoch_action == "restart"
+
+  post_tasks:
+    - name: Prepare health checks
+      include_tasks: tasks/health_check.yml
+      when: epoch_action != "stop"
--- a/ansible/manage-node.yml
+++ b/ansible/manage-node.yml
@@ -1,0 +1,64 @@
+---
+- name: Manage epoch nodes
+  hosts: all
+  remote_user: epoch
+
+  vars:
+    epoch_bin: "{{ ansible_env.HOME }}/node//bin/epoch"
+    epoch_action: "{{ cmd | default('ping') }}"
+    epoch_running: false
+
+  tasks:
+    - name: Fail on unsupported command
+      fail:
+        msg: "Invalid command. Supported commands are: ping, start, stop, restart"
+      when:
+        - epoch_action != "ping"
+        - epoch_action != "start"
+        - epoch_action != "stop"
+        - epoch_action != "restart"
+      connection: local
+
+    - name: Epoch binary exists
+      stat:
+        path: "{{ epoch_bin }}"
+      register: epoch
+      changed_when: false
+      failed_when: epoch.stat.exists != True
+
+    - name: Ping epoch daemon
+      command: "{{ epoch_bin }} ping"
+      register: ping
+      failed_when: no
+      changed_when: false
+      when: epoch.stat.exists == True
+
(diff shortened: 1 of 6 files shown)


ansible/manage-node.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Stop epoch deamon
+      command: "{{ epoch_bin }} stop"
+      when:
+        - epoch_running
+        - epoch_action == "stop"
+
+    - name: Restart epoch deamon
+      command: "{{ epoch_bin }} restart"
+      when:
+        - epoch_running
```
lucafavatella: So if node not running and user issues restart, node stays down without a warning printed? May I suggest deleting restart so user is forced to run stop first then start?

ansible/reset-net.yml
```diff
@@ -0,0 +1,67 @@
...
+      tags: [config]
+
+    - name: Creating new log folder
+      file:
+        path: "{{ project_root }}/log/"
+        state: directory
+      tags: [config]
+
+    - name: Start epoch deamon
+      command: "{{ epoch_bin }} start"
```
lucafavatella: health check?

ansible/manage-node.yml
dincho: This case is handled by always starting the node in https://github.com/aeternity/infrastructure/pull/17/files/8f8e7a237f65aafc928c22b77b00c32d6b68239d#diff-345c1fc9b8d50aad9657a44fa308e014R52

ansible/manage-node.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: Manage epoch nodes
+  hosts: all
+  remote_user: epoch
+
+  vars:
+    epoch_bin: "{{ ansible_env.HOME }}/node//bin/epoch"
```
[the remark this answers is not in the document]
dincho: cleanup in 51b79c1

ansible/manage-node.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: Fail if not running
+      fail:
+        msg: "Ping failed"
+      when:
+        - not epoch_running
+        - epoch_action == "ping"
+
+    - name: Start epoch deamon
+      command: "{{ epoch_bin }} start"
```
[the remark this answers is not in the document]
dincho: Added in 51b79c1

ansible/tasks/health_check.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Wait epoch node API to boot
+  wait_for:
+    port: "{{ config.http.external.port }}"
+    host: "{{ ansible_ssh_host }}"
+    timeout: 30
```
lucafavatella: It is ok - though just for ensuring we are on same page I notice in normal deployment it is more https://github.com/aeternity/epoch/blob/ac7f53b18d612a35ed8575544e162cb4cdbbc68f/deployment/ansible/deploy.yml#L155

dincho: I tested the timeout and and really prefer "fail fast", 30 seconds should be enough to boot up. 5 minutes is away too much.

I know it's different on epoch repo deploy.yml, I'll probably change it there in future too.

P.S. This is one of the reasons I didn't initially added health checks - it's a duplicate of epoch repo and I didn't come up with good solution to this yet.

ansible/reset-net.yml
```diff
@@ -64,6 +75,7 @@
 
     - name: Start epoch deamon
       command: "{{ epoch_bin }} start"
+      when: epoch.stat.exists
```
lucafavatella: If epoch bin does not exist I would rather expect script to have failed and not to arrive at this point. Would it make sense to include explicit fail steps early in the playbook checking for the preconditions? (I would expect script to fail explicitly - rather than skipping tasks - if epoch bin is absent or epoch config is absent.)

dincho: This is just a linter fix/hack.

see https://github.com/aeternity/infrastructure/pull/17/files#diff-c7b2f1005a3b723ea1cfc920d97fea1cR15

PR #18, opened 2018-01-19
Update datadog http_url check URL to v2

dincho:
There is no PT ticket for this. Simple URL update to v2.

Deploy group is also changed because we agreed to add this node to the monitoring some time ago.

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -39,7 +39,7 @@
         init_config:
         instances:
           - name: API
-            url: "http://localhost:3013/v2/top"
+            url: "{{ api_base_uri }}/top"
             # Default is false, i.e. emit events instead of service checks.
             # Recommend to set to true.
             skip_event: true
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -39,7 +39,7 @@
         init_config:
         instances:
           - name: API
-            url: "http://localhost:3013/v1/top"
+            url: "http://localhost:3013/v2/top"
             # Default is false, i.e. emit events instead of service checks.
             # Recommend to set to true.
             skip_event: true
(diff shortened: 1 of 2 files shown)


ansible/monitoring.yml
```diff
@@ -39,7 +39,7 @@
         init_config:
         instances:
           - name: API
-            url: "http://localhost:3013/v1/top"
+            url: "http://localhost:3013/v2/top"
```
michalzajda: should we keep the `vN` in some central place?

dincho: Good point, After we merge #17 I'll rebase and use the variable.

deploy.sh
```diff
@@ -15,5 +15,5 @@ ansible-galaxy install -r requirements.yml
 ansible-playbook environments.yml
 ansible-playbook --limit=epoch setup.yml
 # split monitoring playbook by environment because Ansible group vars are merged otherwise
-ansible-playbook --limit='uat:&epoch:!tester' --extra-vars "datadog_api_key=${datadog_api_key:?}" monitoring.yml
```
lucafavatella: Spurious change?

dincho: I mentioned in the PR description that the tester node is already monitored, this just reflects the current state.

ansible/monitoring.yml
dincho: see dd59135

ansible/monitoring.yml
```diff
@@ -39,7 +39,7 @@
         init_config:
         instances:
           - name: API
-            url: "http://localhost:3013/v2/top"
+            url: "{{ api_base_uri }}/top"
```
lucafavatella: I see defined in https://github.com/aeternity/infrastructure/blob/95c8d0dd616436ad5d8e11b2803d2d787e24b533/ansible/group_vars/epoch/vars.yml#L5

PR #20, opened 2018-02-06
Use Ansible vault for secrets

dincho:
- datadog callback plugin has been updated to be able to use datadog_api_key, however still sux because it can be defined only in the inventory (encrypted in this case)
- Travis environment variable ANSIBLE_VAULT_PASSWORD is configured

[PT 154166551](https://www.pivotaltracker.com/story/show/154166551)


ansible/vault-env
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo $ANSIBLE_VAULT_PASSWORD
```
lucafavatella: I understand this is a custom variable - not relevant for ansible (the name is similar to names used by ansible for its env vars.)

ansible/ansible.cfg
```diff
@@ -1,8 +1,9 @@
 [defaults]
-callback_whitelist = profile_tasks
+callback_whitelist = profile_tasks,datadog_callback
 host_key_checking = False
-inventory = inventory/openstack.yml
+inventory = inventory
 timeout = 30
+vault_password_file = vault-env
```
lucafavatella: I assume that for some Ansible-specific general rule `vault_password_file` var here is equivalent to `--vault-password-file` from CLI.

Based on https://docs.ansible.com/ansible/2.4/vault.html#providing-vault-passwords
* The recommended options is `--vault-id` (rather than `--vault-password-file`);
* I see `vault-env` is a "a vault password executable script" though this capability of being a script, whose stdout is relevant, rather than a plain file, whose content is relevant, is documented for `--vault-id` - not for `--vault-password-file`.

ansible/vault-env
dincho: indeed, it's the env var we'll use to configure the password

dincho:
> Does Circle CI have similar capability? See https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings - malicious user may attempt to echo/print/send var by opening PR (or forking).

Yes, it does. [more details in the docs](https://circleci.com/docs/1.0/fork-pr-builds/)

@sennui please make sure this is turned off for our repos (I believe it's off by default, but let's double check)

ansible/ansible.cfg
dincho: Well, ansible.cfg options are documented, while it may looks like there 1:1 match with the cli options, I don't think it's the case.

I see only [`vault_password_file` documented](docs.ansible.com/ansible/2.4/intro_configuration.html#vault-password-file) and no `vault_id` or related.

lucafavatella: Ah great thanks. And documented to be potentially a script. Thanks.

PR #21, opened 2018-02-14
Install libsodium to all nodes

dincho:


ansible/setup.yml
```diff
@@ -70,3 +72,26 @@
...
       tags:
         - dev-tools
+
+    - name: Download libsodium
+      unarchive:
+        src: https://github.com/jedisct1/libsodium/releases/download/{{ libsodium_version }}/libsodium-{{ libsodium_version }}.tar.gz
+        dest: /usr/local/src/
+        remote_src: yes
+      args:
+        creates: /usr/local/src/libsodium-{{ libsodium_version }}.tar.gz
```
lucafavatella: I am surprised unarchive creates a .tar.gz

ansible/setup.yml
```diff
@@ -19,6 +19,8 @@
       - happi
       - UlfNorell
       - gorillainduction
+    libsodium_version: "1.0.16"
```
lucafavatella: I see https://github.com/aeternity/epoch/blob/66ea4da55bb919eac10985c53ac920b46fba6cc3/docker/Dockerfile-builder#L22-L28
Good

ansible/setup.yml
lucafavatella: From the documentation at http://docs.ansible.com/ansible/latest/unarchive_module.html this is still unclear to me

dincho: This was a bug/obsolete configuration fixed in c14dfe6, the module is already idempotent (checks if dest: dir exists).

dincho: good catch btw

PR #22, opened 2018-02-14
Fix travis deploys

dincho:


dincho:
this is obsolete, we're migrating to CircleCI

PR #24, opened 2018-02-17
CircleCI

dincho:
Migrate this repo to CircleCI. Temporary disable UAT infrastructure setup, until it's confirmed it works as intended.

Also add automatic daily run (12:00 UTC), so we can catch configuration issues on time, instead of otherwise rarely runs/updates.

[Example successful run.](https://circleci.com/gh/aeternity/infrastructure/79)


ansible/environments.yml
```diff
@@ -17,6 +17,8 @@
 #| m1.xlarge | 16384 |   40 |     8 |
 #+-----------+-------+------+-------+
 
+# Using Image ID d324da4b-b827-44fb-8331-bfad41a28aca instead of Ubuntu 16.04.3,
```
lucafavatella: Could you please clarify where this (copied below) string comes from?

lucafavatella: I am sure it shall be derived from https://docs.openstack.org/image-guide/obtain-images.html#ubuntu though not sure how.

ansible/setup.yml
```diff
@@ -40,6 +40,18 @@
         - ssh-keys
         - devops-ssh-keys
 
+    # Automatic cleanup is not supported. You must remove old keys manually.
+    - name: "Add CI master keys"
+      authorized_key:
+        user: "{{ ansible_user }}"
+        key: "{{ lookup('file', item) }}"
+      with_items:
+        - files/travis/master_key.pub
```
lucafavatella: Not sure where this file comes from.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      - checkout
+      - *restore_venv
+      - *activate_venv
+      - *restore_roles
+      # Don't run on UAT until we're sure it won't cause changes
+      - run:
+          name: Setup infrastructure
+          command: |
+            cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python"' \
+              --skip-tags "uat-env" environments.yml
```
lucafavatella: I see
https://github.com/aeternity/infrastructure/blob/5308db4e90e8a73ad27162230b3e9097568fb885/ansible/environments.yml#L57
http://manpages.ubuntu.com/manpages/xenial/en/man1/ansible-playbook.1.html
so shall be ok. Though still keep an eye on envs when merging.

lucafavatella:
> Example successful run.

Is
```
#!/bin/bash -eo pipefail
cd ansible && ansible-playbook --limit='epoch' setup.yml


PLAY [Configure node servers] **************************************************
Could not load configuration, invalid file: /home/builder/infrastructure/ansible/callback_plugins/datadog_callback.yml
```
expected?

ansible/setup.yml
lucafavatella: I understood - it is relative path so https://github.com/aeternity/infrastructure/blob/97bf5d80f6a739eefdc039fad5c83fe421a79ca9/ansible/files/travis/deploy_key.pub

Shall these travis-related files be deleted too?

dincho:
> Example successful run.
> Is
> 
> #!/bin/bash -eo pipefail
> cd ansible && ansible-playbook --limit='epoch' setup.yml
> 
> 
> PLAY [Configure node servers] **************************************************
> Could not load configuration, invalid file: /home/builder/infrastructure/ansible/callback_plugins/datadog_callback.yml
> expected?

Yes, unfortunately. This is how the callback plugin works, it looks for configuration file and [warn if not found](https://github.com/aeternity/infrastructure/blob/master/ansible/callback_plugins/datadog_callback.py#L62), however we're using an [inventory variables](https://github.com/aeternity/infrastructure/blob/master/ansible/callback_plugins/datadog_callback.py#L249) defined [in the vault](https://github.com/aeternity/infrastructure/blob/master/ansible/inventory/vault.yml)

(let's me know if you're interested to see the contents of the vault, I'll share the key with you)

ansible/environments.yml
dincho: One can find the ID by listing all private images:

```
$ openstack image list --private
+--------------------------------------+----------------+--------+
| ID                                   | Name           | Status |
+--------------------------------------+----------------+--------+
| 7fbe9a7c-05e3-4a80-a59f-682fef4d311e | Ubuntu 14.04.5 | active |
| d324da4b-b827-44fb-8331-bfad41a28aca | Ubuntu 16.04.3 | active |
+--------------------------------------+----------------+--------+
```

(one could also use the UI of course).

ansible/setup.yml
dincho: Nice catch, I deleted all travis keys in dd06a4e

PR #27, opened 2018-03-13
Change inventory groups naming for compatibility

dincho:
AWS dynamic inventory plugin is using the pattern "tag_{key}_{value}"

This change would allow us to use the same tags for both AWS and OS.
The change is backward compatible, old tags are also kept (for now).

in context of [PT 155784885](https://www.pivotaltracker.com/story/show/155784885)


ansible/environments.yml
```diff
@@ -33,7 +33,7 @@
         template: "../openstack/ae-environment.yml"
         parameters:
           env_name: integration
-          meta_groups: "epoch,integration"
+          meta_groups: "epoch,integration,tag_role_epoch,tag_env_integration"
```
lucafavatella: I understand that keeping these old tags is meant fo backwards compatibility of scripts.
Though I see that group_vars are moved - not copied: do I understand correctly that if scripts called with old tags then some vars will not be loaded and some scripts will not work?
Not sure what benefit is of keeping old tags - if none they could be removed for simplicity.

dincho:
> Shall
> 
> infrastructure/openstack/test/create.yml
> Line 3 in db26613
>  meta_groups: "epoch,uat" 
> be updated too?

This is a test file, does not make much sense to update it.

ansible/environments.yml
dincho: The only reason for this BC is because of https://github.com/aeternity/epoch/pull/856 that should be merged after this.

This repository does not need BC, because well, as you noticed the group_vars are renamed.

And once https://github.com/aeternity/epoch/pull/856 is merged, we can remove this BC.

PR #30, opened 2018-03-20
setup ulimit

acidsennui:
@dincho

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -86,7 +86,7 @@
 
     - name: Increase ulimit
       pam_limits:
-        domain: {{ project_user }}
+        domain: "{{ project_user }}"
         limit_type: soft
         limit_item: nofile
         value: 24576
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -84,6 +84,13 @@
       tags:
         - dev-tools
 
+    - name: Increase ulimit
+      pam_limits:
+        domain: epoch
+        limit_type: soft
+        limit_item: nofile
+        value: 24576
+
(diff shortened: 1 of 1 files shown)


dincho:
context: https://www.pivotaltracker.com/story/show/155932652

ansible/setup.yml
```diff
@@ -84,6 +84,13 @@
       tags:
         - dev-tools
 
+    - name: Increase ulimit
+      pam_limits:
+        domain: epoch
```
lucafavatella: Shall this rather be `"{{ project_user }}"`?

ansible/setup.yml
```diff
@@ -84,6 +84,13 @@
...
+        limit_type: soft
+        limit_item: nofile
+        value: 24576
```
lucafavatella: I see this is taken from https://github.com/aeternity/epoch/blob/v0.9.0/hooks/pre_start.sh#L3
Ok.

ansible/setup.yml
```diff
@@ -84,6 +84,13 @@
       tags:
         - dev-tools
 
+    - name: Increase ulimit
+      pam_limits:
+        domain: epoch
+        limit_type: soft
```
lucafavatella: Can you please confirm that this step fails if hard limit is lower so this soft limit is ineffective?

ansible/setup.yml
acidsennui: good catch

ansible/setup.yml
acidsennui: yes

PR #35, opened 2018-03-26
More nodes in Europe and Asia

acidsennui:
create 2 additional 10 nodes envs in Europe and Asia

PT: https://www.pivotaltracker.com/story/show/153367499 
PT: https://www.pivotaltracker.com/story/show/153367769

### DIFF
--- a/ansible/environments.yml
+++ b/ansible/environments.yml
@@ -118,3 +118,33 @@
           NodeCount: 10
           AmiId: "ami-34a9364c" # AMI  ubuntu 16.04.3 with master user + python added.
           InstanceType: "m4.large"
+
+    - name: Ensure ae-uat-eucentral1 environment stack
+      cloudformation:
+        stack_name: "ae-uat"
+        state: "present"
+        region: "eu-central-1"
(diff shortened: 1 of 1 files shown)


ansible/environments.yml
```diff
@@ -118,3 +118,33 @@
           NodeCount: 10
           AmiId: "ami-34a9364c" # AMI  ubuntu 16.04.3 with master user + python added.
           InstanceType: "m4.large"
+
+    - name: Ensure ae-uat-eucentral1 environment stack
+      cloudformation:
+        stack_name: "ae-uat"
```
lucafavatella: I see ae-uat used in 20+10+ (not in this PR) 10+10 nodes.
Are they connected automatically as peers in the config of the application running in UAT somehow? And what about running nodes?

acidsennui: 10 new in eu-central-1 10, new in ap-southeast-1, and 10 already created in us-west-2

dincho: And 20 more at eu-east (OS Sofia)

PR #36, opened 2018-03-26
create static ip

acidsennui:
ref: https://www.pivotaltracker.com/story/show/155658506

so i can create ElasticIP (static ip) in cloud formation template, but there is now way that i will be able to assign it to one of created instances.

there is no way to get instance IP from CF template when creating instances with autoscaling group.

But after creating IP manual assignign this IP do not broke anything in CF template.

### DIFF
--- a/aws/ae-environment.yml
+++ b/aws/ae-environment.yml
@@ -97,3 +97,6 @@ Resources:
         AvailabilityZones:
           Fn::GetAZs:
             Ref: "AWS::Region"
+
+    EpochElasticIP:
+      Type: "AWS::EC2::EIP"


dincho:
This looks weird to me. What IP to which node will be assigned after this configuration change?

acidsennui:
this will change only if we will recreate stack. If you would create instances "one by one" in CloudFormation template instead of Auto Scaling Group, i would be able to assing IP to one of instances. There is just now way in CloudFormation tempalte to get instance id from ASG ;/
 
So i can change it to create 10 nodes there not in ASG and this would solve this issue and also naming issue.

dincho:
> So i can change it to create 10 nodes that are not in ASG and this would solve this issue and also naming issue.

Does this means that you have to copy-paste 10 rows of per instance configuration ?

dincho:
> This looks weird to me. What IP to which node will be assigned after this configuration change?

Oh, so it just creates a elastic IP .. no assignment. I see.

acidsennui:
yes, solution would be copy & paste 10 rows for instances.

acidsennui:
which is stupid, horrible and make be dizy :D but this would solve naming issue, and static ip issue.

aws/ae-environment.yml
```diff
@@ -97,3 +97,6 @@ Resources:
         AvailabilityZones:
           Fn::GetAZs:
             Ref: "AWS::Region"
+
+    EpochElasticIP:
+      Type: "AWS::EC2::EIP"
```
lucafavatella: I see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-eip.html

PR #38, opened 2018-03-29
Delete old keys during net resets

dincho:


dincho:
> Would this not reset the pub-priv keypair of testnet peers documented in user docs?

This is "manual" playbook, that we sometimes run to cleanup and environment. It's not excepted to be run on the testnet, at least until it we have deployed static keys.

PR #39, opened 2018-04-04
remove IP from CloudFormation tempalte

acidsennui:


aws/ae-environment.yml
```diff
@@ -97,6 +97,3 @@ Resources:
         AvailabilityZones:
           Fn::GetAZs:
             Ref: "AWS::Region"
-
-    EpochElasticIP:
```
lucafavatella: So the idea is removing this because unused, right?

lucafavatella: Would removing this section stop billing of the ElasticIP? Do I understand correctly that manual operation is needed for actually deleting ElasticIP? Did you plan for those operations?

dincho: The idea of removing this is otherwise deleting the stack will delete the elastic IP too, we don't want that, but to keep it "forever".

dincho: Already deleted.

PR #40, opened 2018-04-05
Add script for dumping ansible inventory to MD

dincho:
As part of https://www.pivotaltracker.com/story/show/156536732 I created small python script intended for manual usage.

This may evolve to automatic task that updates the datadog dashboard.


dincho:
> I see this as the main value of this PR.

Considering that the infrastructure nodes are changed very rarely I consider this additional work not needed yet.

PR #44, opened 2018-04-17
Add platform tag accross all envs

acidsennui:
PT: https://www.pivotaltracker.com/story/show/156435960

added platform tag accross all envs. not nice way for AWS but no other way with CloudFormation

### DIFF
--- a/ansible/environments.yml
+++ b/ansible/environments.yml
@@ -35,7 +35,7 @@
         template: "../openstack/ae-environment.yml"
         parameters:
           env_name: integration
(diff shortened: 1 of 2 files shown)


dincho:
> Both ticket and PR say what this does but not why: I do not understand why needed or what problem solved.

Your comment made me rethink this and it's implementation is totally wrong.

acidsennui:
we need those tags as datadog tags.

dincho:
see https://github.com/aeternity/infrastructure/pull/50

PR #46, opened 2018-04-19
Cleanup old environment tags

dincho:
https://www.pivotaltracker.com/story/show/156435936


dincho:
@lucafavatella see https://github.com/aeternity/epoch/pull/1037

PR #47, opened 2018-04-24
setup parallelism: 4

acidsennui:
pt: https://www.pivotaltracker.com/story/show/156844586

according to docs: https://circleci.com/docs/2.0/parallelism-faster-jobs/#specifying-a-jobs-parallelism-level this should work

cc @dincho


acidsennui:
@dincho not sure i understand ... how to shuld be solved

PR #48, opened 2018-04-25
Pt 156378851 region and platform tags

acidsennui:
pt: https://www.pivotaltracker.com/story/show/156378851


acidsennui:
those should be datadog tags. ;]

PR #49, opened 2018-04-26
Pt 156838514 aws hostname fix

acidsennui:
PT: https://www.pivotaltracker.com/story/show/156838514

This will not affect any running isntances, this will just update LC in ASG, no other actions will be taken, all new instances will have fixed hostname.
To have it fixed on old hosts, would have to rotate all instances to do this and persist static ip

1. Change manualy ASG to count + 1, static ip to new instance.
2. Change ASG count to count +9, wait for new instances to boot
3. Change ASG count back to 10, oldest instances will be killed.

### DIFF
--- a/aws/ae-environment.yml
+++ b/aws/ae-environment.yml
@@ -87,6 +87,8 @@ Resources:
     EpochServerAutoScalingGroup:
       Type: "AWS::AutoScaling::AutoScalingGroup"
       Properties:
+        TerminationPolicies:
+          - OldestInstance
         AutoScalingGroupName:
           Fn::Sub:
             - "ae-${EnvName}-auto-scaling-group"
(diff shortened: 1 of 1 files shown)


acidsennui:
Info about `magic ip address` there https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

dincho:
@acidsennui can we also change the instance (tag) name? Or it does not makes sense 🤔

dincho:
> To have it fixed on old hosts, would have to rotate all instances to do this and persist static ip
> 
> Change manualy ASG to count + 1, static ip to new instance.
> Change ASG count to count +9, wait for new instances to boot
> Change ASG count back to 10, oldest instances will be killed.

Or we can just exec this piece of bash to all running instances?

acidsennui:
@dincho yes, instance can update it's tag also on boot to have this name there.

yes also we can just exec script to fix hostnames, just not sure how datadog and epoch will react on this change?

dincho:
> Both ticket and PR mention fix though not sure of symptom.

The default is the AWS instance ID which is not convenient when you login.

PR #50, opened 2018-04-30
add tags for platform and env

acidsennui:
PT: https://www.pivotaltracker.com/story/show/156378851


dincho:
> Who assigns the nodes to one (or more?) groups? In particular, does "regionOne" come from a non-published nova/openstack related config?

It comes from the [dynamic inventory script](https://github.com/ansible/ansible/blob/stable-2.5/lib/ansible/plugins/inventory/openstack.py#L258) the name itself is set by the OS provider.

PR #51, opened 2018-05-01
Polish environments playbook tags

dincho:
https://www.pivotaltracker.com/story/show/156435960

### DIFF
--- a/ansible/environments.yml
+++ b/ansible/environments.yml
@@ -40,7 +40,8 @@
           node_flavor: c1.medium
           image_name: "d324da4b-b827-44fb-8331-bfad41a28aca"
       tags:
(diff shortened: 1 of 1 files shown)


ansible/environments.yml
```diff
@@ -40,7 +40,8 @@
           node_flavor: c1.medium
           image_name: "d324da4b-b827-44fb-8331-bfad41a28aca"
       tags:
-        - integration-env
```
lucafavatella: I did not check the `-env` suffix.

dincho: I don't find any usage, it's primary purpose was manual runs. I changed it for consistency - "the ansible way" of tags (no names, only values).

PR #52, opened 2018-05-01
Split environments setup

dincho:
This will reduce the pipeline runtime to [about 30 minutes](https://circleci.com/workflow-run/d42baf84-dd54-4cff-bc5b-3ccce3316c5d) (longest job pipeline being UAT).

Otherwise it runs for [about 35 minutes](https://circleci.com/workflow-run/5a1eb5a7-0313-42ea-b0e2-631b9e82012a)

https://www.pivotaltracker.com/story/show/156844586

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,25 @@ references:
         source ~/infrastructure/venv/ansible/bin/activate
         echo "source ~/infrastructure/venv/ansible/bin/activate" >> $BASH_ENV
 
+  setup_infra: &setup_infra
+    run:
+      name: Setup infrastructure
+      command: |
+        cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python"' \
+          --tags "${DEPLOY_ENV:?}" environments.yml
+
+  setup_node: &setup_node
+    run:
+      name: Setup environment
+      command: |
+        cd ansible && ansible-playbook --limit="tag_env_${DEPLOY_ENV:?}:&tag_role_epoch" setup.yml
+
+  setup_monitoring: &setup_monitoring
+    run:
+      name: Install monitoring tools
+      command: |
+        cd ansible && ansible-playbook --limit="tag_env_${DEPLOY_ENV:?}:&tag_role_epoch" monitoring.yml
+
 jobs:
   build:
     <<: *container_config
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -79,36 +98,69 @@ jobs:
...
+      - *activate_venv
+      - *restore_roles
+      - *setup_infra
+      - *setup_node
+      - *setup_monitoring
+
+  setup_fast_integration:
+    <<: *container_config
+    environment:
+      - DEPLOY_ENV: fast_integration
```
lucafavatella: I see labels as in https://github.com/aeternity/infrastructure/blob/56cfd6326b59f8e91afe53397b7c1d2f5ae11eae/ansible/environments.yml#L113 ok

.circleci/config.yml
```diff
@@ -79,36 +98,69 @@ jobs:
             cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python"' \
               --check -i localhost, environments.yml
 
-  deploy:
+  setup_uat:
     <<: *container_config
+    environment:
+      - DEPLOY_ENV: uat
```
lucafavatella: I see some traces of static_uat https://github.com/aeternity/infrastructure/blob/56cfd6326b59f8e91afe53397b7c1d2f5ae11eae/ansible/dump_inventory.py#L16
Does this interact with that in any way?

.circleci/config.yml
```diff
@@ -31,6 +31,25 @@ references:
...
+      name: Setup infrastructure
+      command: |
+        cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python"' \
+          --tags "${DEPLOY_ENV:?}" environments.yml
+
+  setup_node: &setup_node
+    run:
+      name: Setup environment
+      command: |
+        cd ansible && ansible-playbook --limit="tag_env_${DEPLOY_ENV:?}:&tag_role_epoch" setup.yml
```
lucafavatella: I see [ `:&` in host pattern is intersection (while simple `:` in host pattern is OR)](https://docs.ansible.com/ansible/2.4/intro_patterns.html).

I see labels tag_role_epoch and tag_env* (e.g. tag_env_uat) are assigned to hosts in https://github.com/aeternity/infrastructure/blob/56cfd6326b59f8e91afe53397b7c1d2f5ae11eae/ansible/environments.yml#L55 though it is not clear to me how meta_group works.

What concerns me is that I do not see these labels assigned to AWS nodes https://github.com/aeternity/infrastructure/blob/56cfd6326b59f8e91afe53397b7c1d2f5ae11eae/ansible/environments.yml#L117-L169 (and I do not see dump_inventory called anywhere by CI) hence AWS nodes excluded. Correct? How does this process extend to AWS nodes from CI then?

.circleci/config.yml
```diff
@@ -137,4 +217,8 @@ workflows:
               only:
                 - master
     jobs:
-      - deploy
+      - setup_uat
```
lucafavatella: I see this full parallelization enables reducing time. Though I am concerned that, if master has a bug detectable by step `setup_node` or earlier step, that bug would be executed on all environments.
* Have you considering serializing most envs to earlier setup of one of the environments e.g. setup_dev1? That would slow down CI though increase confidence.
* In alternative, could you please confirm how PR gets checked before merge to master? `check` job seems performing basic sanity check though not the full setup on a dummy env.

dincho: This is actually another feature (besides the speedup), failing one environment should not fail the others.

Most of the times the code has been run and tested manually/locally. I'm 👍 to add additional environment to run actually playbooks on it. However IMO this is not in scope of this PR.

.circleci/config.yml
dincho: We setup the same AWS tags in https://github.com/aeternity/infrastructure/blob/master/aws/ae-environment.yml#L99-L117

.circleci/config.yml
dincho: This is leftover, not really related but I removed it in 8affbf0

PR #56, opened 2018-05-14
(re)Enable UAT infrastructure setup

dincho:
While the changes has nothing to do with a fix, I fixed the problem manually so that UAT auto infrastructure setup can be re-enabled.

https://www.pivotaltracker.com/story/show/157480213

### DIFF
--- a/ansible/environments.yml
+++ b/ansible/environments.yml
@@ -62,6 +62,7 @@
       tags:
         - uat
         - os
+        - os-uat
 
     - name: Ensure dev1 environment stack
       os_stack:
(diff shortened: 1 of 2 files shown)


ansible/environments.yml
```diff
@@ -62,6 +62,7 @@
       tags:
         - uat
         - os
+        - os-uat
```
lucafavatella: This seems inconsistent with tags in other envs? Where is this used? Why is it needed?

dincho: Yes, it looks inconsistent. The problem it solves is that ansible cannot "combine" tags. Without this tag we can't run tasks for UAT on OS only. We don't have other environments on AWS, thus the inconsistency.

PR #57, opened 2018-05-16
Update datadog agent configuration to v6 syntax

dincho:
- effectively enabling process monitoring

This has been tested on fast integration n0, both manually and with playbook run.

https://www.pivotaltracker.com/story/show/156378779

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -11,8 +11,9 @@
     datadog_config:
       apm_enabled: false
       use_dogstatsd: true
-      process_agent_enabled: true
-      log_enabled: true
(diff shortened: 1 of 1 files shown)


ansible/monitoring.yml
```diff
@@ -11,8 +11,9 @@
     datadog_config:
       apm_enabled: false
       use_dogstatsd: true
-      process_agent_enabled: true
-      log_enabled: true
+      process_config:
+        enabled: "true" # has to be set as a string
+      logs_enabled: true
```
lucafavatella: I guess the added s is intended.

PR #61, opened 2018-06-07
Add Packer configuration

dincho:
Image provisioners install everything we would install otherwise too. It also updates the distros.
It does not keep datadog running and do not deploy it's API key, also some of it's tags are set as unknown. This means that some runtime provisioning is still required, but it should be much faster now.

Building images is meant to be triggered/run manually, either locally or pushing to `packer` branch.

Example run: https://circleci.com/gh/aeternity/infrastructure/752


packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+      "source_ami_filter": {
+        "filters": {
+          "virtualization-type": "hvm",
+          "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477"],
+        "most_recent": true
+      },
+      "instance_type": "t2.micro",
```
dincho: We don't need too much faster builds, so running cheap build instance.

packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+    {
+      "type": "amazon-ebs",
+      "region": "us-west-2",
+      "source_ami_filter": {
+        "filters": {
+          "virtualization-type": "hvm",
+          "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477"],
```
dincho: Canonical account.

packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+        "most_recent": true
+      },
+      "instance_type": "t2.micro",
+      "ssh_username": "ubuntu",
+      "ami_name": "epoch-ubuntu-16.04-v{{timestamp}}",
+      "ami_regions": ["eu-central-1", "ap-southeast-1"],
```
dincho: Additional copy to those 2 regions, total of 3 (see above `region`)

packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+          "virtualization-type": "hvm",
+          "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477"],
+        "most_recent": true
+      },
+      "instance_type": "t2.micro",
+      "ssh_username": "ubuntu",
+      "ami_name": "epoch-ubuntu-16.04-v{{timestamp}}",
```
dincho: While AWS builder supports `force_deregister`, OS does not. Thus `{{timestamp}}` is used here for consistency.
Otherwise I'd prefer manual changing v1, v2, vX and don't waster time to cleanup old images manually.

packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+      "ami_regions": ["eu-central-1", "ap-southeast-1"],
+      "force_deregister": true,
+      "force_delete_snapshot": true
+    },
+    {
+      "type": "openstack",
+      "ssh_username": "master",
+      "image_name": "epoch-ubuntu-16.04-v{{timestamp}}",
+      "source_image": "5566886d-0d9d-4d86-94ff-55597169e15a",
```
dincho: The build in provider Ubuntu 16.04 image, shall we use 16.04.3 (also build by the OS provider) to speed the build up?

packer/epoch.json
```diff
@@ -0,0 +1,74 @@
...
+      "flavor": "c1.tiny",
+      "networks": ["a07d7f52-0eed-4242-91fd-368dea025cc7"]
```
dincho: ID of the external (public) network.

packer/epoch.json
dincho: The OS provider image 16.04.3 is too old, I've used our own private 16.04.4 image in 80603c8d024281a626c19ab0081f6007efd7d418

dincho: New image with fixed cloud-init: c923ab5df6ac795aec6850e17140a9740a638f65

.circleci/config.yml
```diff
@@ -198,6 +217,11 @@ workflows:
             branches:
               only: master
 
+      - build_images:
+          filters:
+            branches:
+              only: packer
```
lucafavatella: Could you please mark from github settings branch `packer` as protected somehow? E.g. so that no one can delete branch from https://github.com/aeternity/infrastructure/branches

packer/apt-upgrade.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+while sudo fuser -s /var/lib/dpkg/lock; do
```
lucafavatella: Can you please point me to the documentation of the exit status of this command? Cases: file not present, file present with no process having it open, file present with at lease a process having it open.

.circleci/config.yml
lucafavatella: I see `packer` branch is same as this PR i.e. 367aadd

dincho: @sennui Could we please protect `packer` branch ?

packer/apt-upgrade.sh
dincho: I'm afraid it's not documented. I tested it myself. However I also found this for you: https://gitlab.com/psmisc/psmisc/blob/master/src/fuser.c#L1355

I also assume that the command will return non 0 exit code if the file does not exists, however we can assume it always exists because of base image usage (where it's always present).

ansible/image.yml
```diff
@@ -0,0 +1,19 @@
+---
+- import_playbook: setup.yml
```
lucafavatella: It seems the scripts setup.yml and monitoring.yml are used both:
* When creating images (this PR - packer branch only);
* At midday every day - on each host - from master branch (see [master](https://github.com/aeternity/infrastructure/blob/6f100077951b7f73384f0a90a6eeb63708c5d145/.circleci/config.yml#L204)).

So these scripts have double use.

I am concerned re how the two branches (master and packer) will be updated when these scripts (setup.yml and monitoring.yml) change. What workflow did you have in mind?

dincho:
> Skimming this PR it seems to me that images are built using packer though usage of such images is not in this PR. Or am I missing it?

Correct. I don't think we need to change our current node images. I assume usage of the new images only for new environments (or if we need to rebuild current environments).

ansible/image.yml
dincho: The workflow that I have in mind is to run `packer` manually (pushing to remote branch or locally).

The `packer` build images is just a base, that would speedup instances initial provisioning.

Then the daily (or master) run of the playbooks keep it in sync with latest changes.


The workflow is the same as we have now, with a different base instance image which will be setup in advance - "build" time instead of "provisioning" time.

lucafavatella: Maybe the [SSH keys setup](https://github.com/aeternity/infrastructure/blob/6f100077951b7f73384f0a90a6eeb63708c5d145/ansible/setup.yml#L9-L23) shall be moved out of the AMI creation?

dincho: So, in conclusion - the packer base image does not need to be "very" up-to-date, of course we'd keep it, but even it's forgotten it's not a problem.

dincho: What's the reason?

This is one of the slowest tasks ATM.

lucafavatella: So it looks like the problem addressed and the workflow is:

* If DevOps needs to instantiate env with more than e.g. 3 nodes, prefer using recent packer-based images (AWS, openstack).
* In order to refresh the package-based images (AWS, openstack), force-push recent `master` branch to `packer` branch.

Right? If yes, `packer` branch shall *not* be protected (as `env/dev1` in epoch repo).

Could you please README.md ?

lucafavatella: Ok. Please refer to https://github.com/aeternity/infrastructure/pull/61#discussion_r194372872

.circleci/config.yml
dincho: Rethinking, deletion of the branch looks "harmful" to me? The next time someone push it, it will resurrect and continue working ?

If we protect it, it will also disable force pushing to it, which is better IMO than merging for this kind of branches (it's just a trigger).

lucafavatella: [Telepathy](https://github.com/aeternity/infrastructure/pull/61#discussion_r194372872).

ansible/image.yml
dincho: yep, see b161e50

PR #62, opened 2018-06-13
Split UAT to blue/green colors

dincho:
In scope of https://www.pivotaltracker.com/story/show/157217781

Green: 20 eu-east + 10 us-west = 30 nodes
Blue: 10 eu-central + 10 ap-southeast =  20 nodes

### DIFF
--- a/ansible/environments.yml
+++ b/ansible/environments.yml
@@ -53,7 +53,7 @@
         template: "../openstack/ae-environment.yml"
         parameters:
           env_name: uat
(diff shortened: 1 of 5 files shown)


dincho:
> Maybe https://github.com/aeternity/epoch/blame/c277300327ff865ff9cdcd19c53e2edf74c010f6/docs/operation.md#L82 shall be integrated with one IP from the other group? (In case for any reason either one of the blue or the green portion of UAT were not consistent for more than a couple of minutes?)

Good idea, will do.

PR #63, opened 2018-06-15
Pt 157219023 terraform

acidsennui:

PR #64, opened 2018-07-03
Add auto install of epoch to cloud images

acidsennui:
pt: https://www.pivotaltracker.com/story/show/156532207

using user script + AWS KMS to secure inject datadog api key into instance.

TODO:
- [x] run AWS packer build from this branch
- [x] update AMI versions
- [x] a sentence in the docs about Terraform
- [x] rebase to master
- [x] run terraform apply region by region

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,11 @@ references:
       name: Setup infrastructure
       command: make setup-infrastructure
 
+  setup_terraform: &setup_terraform
+    run:
+      name: Setup Terraform
+      command: make setup-terraform
+
   setup_node: &setup_node
     run:
       name: Setup node
@@ -61,6 +66,11 @@ jobs:
       - run:
           name: Dry run environments playbooks
           command: .circleci/docker-run.sh make test-setup-environments
+
+  setup_terraform:
+    <<: *container_config
+    steps:
+      - *setup_terraform
 
   setup_uat:
     <<: *container_config
@@ -132,9 +142,17 @@ workflows:
               ignore:
                 - packer
 
+      - setup_terraform:
+          requires:
+            - build_check
+          filters:
+            branches:
+              only: master
+
       - setup_uat:
           requires:
             - build_check
+            - setup_terraform
           filters:
             branches:
               only: master
@@ -188,7 +206,10 @@ workflows:
               only:
                 - master
     jobs:
-      - setup_uat
+      - setup_terraform
+      - setup_uat:
+          requires:
+            - setup-terraform
       - setup_integration
       - setup_fast_integration
       - setup_dev1
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ images:
 
 setup-infrastructure: check-deploy-env
 	cd ansible && ansible-playbook --tags "$(DEPLOY_ENV)" environments.yml
+
+setup-terraform:
 	cd terraform && terraform init && terraform apply --auto-approve
 
 setup-node: check-deploy-env
@@ -74,6 +76,6 @@ ifndef DEPLOY_ENV
 endif
 
 .PHONY: \
-	images setup-infrastructure setup-node setup-monitoring setup \
+	images setup-infrastructure setup-terraform setup-node setup-monitoring setup \
 	manage-node reset-net lint test-openstack test-setup-environments \
 	check-deploy-env
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -34,9 +34,9 @@
         - project-user
 
     # Automatic cleanup is not supported. You must remove old keys manually.
-    - name: "Setup root SSH authorized keys"
+    - name: "Setup master SSH authorized keys"
       authorized_key:
-        user: "{{ ansible_user }}"
+        user: master
         key: https://github.com/{{ item }}.keys
       with_items: "{{ devops }}"
       tags:
@@ -46,7 +46,7 @@
     # Automatic cleanup is not supported. You must remove old keys manually.
     - name: "Add CI master keys"
       authorized_key:
-        user: "{{ ansible_user }}"
+        user: master
         key: "{{ lookup('file', item) }}"
       with_items:
         - files/circleci/master_key.pub
@@ -78,11 +78,22 @@
     - name: Install dev tools
       apt:
         pkg: "{{ item }}"
-        state: installed
+        state: present
         update_cache: yes
       with_items:
         - build-essential
         - python-yaml
+        - python-pip
+        - python3-pip
+        - jq
+      tags:
+        - dev-tools
+
+    - name: Install pip tools
+      pip:
+        name: "{{ item }}"
+      with_items:
+        - awscli
       tags:
         - dev-tools
 
--- a/packer/epoch.json
+++ b/packer/epoch.json
@@ -52,7 +52,7 @@
     },
     {
       "type": "ansible",
-      "user": "master",
+      "user": "ubuntu",
       "playbook_file": "{{ template_dir }}/../ansible/image.yml",
       "extra_arguments": [
         "-e ansible_python_interpreter='/usr/bin/env python3' datadog_platform=aws"
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -37,6 +37,13 @@ module "aws_deploy-ap-southeast-1" {
   spot_price    = "0.125"
   instance_type = "m4.large"
 
+  epoch = {
+    version = "0.18.0"
+
+    # The string must be bash escaped
+    beneficiary = "ak\\$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
+  }
+
   providers = {
     aws = "aws.ap-southeast-1"
   }
@@ -53,6 +60,13 @@ module "aws_deploy-eu-central-1" {
   spot_price    = "0.125"
   instance_type = "m4.large"
 
+  epoch = {
+    version = "0.18.0"
+
+    # The string must be bash escaped
+    beneficiary = "ak\\$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
+  }
+
   providers = {
     aws = "aws.eu-central-1"
   }
@@ -68,6 +82,13 @@ module "aws_deploy-us-west-2" {
 
   spot_price    = "0.125"
   instance_type = "m4.large"
+
+  epoch = {
+    version = "0.18.0"
+
+    # The string must be bash escaped
+    beneficiary = "ak\\$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
+  }
 
   providers = {
     aws = "aws.us-west-2"
--- a/terraform/modules/cloud/aws/deploy/fleet/iam.tf
+++ b/terraform/modules/cloud/aws/deploy/fleet/iam.tf
@@ -1,0 +1,45 @@
+resource "aws_iam_role" "epoch" {
+  name = "${data.aws_region.current.name}-${var.env}-epoch"
+
+  assume_role_policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "ec2.amazonaws.com"
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_iam_instance_profile" "epoch" {
+  name = "${data.aws_region.current.name}-${var.env}-epoch"
+  role = "${aws_iam_role.epoch.name}"
+}
+
+resource "aws_iam_role_policy" "epoch_policy" {
+  name = "${data.aws_region.current.name}-${var.env}-epoch"
+  role = "${aws_iam_role.epoch.id}"
+
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "secretsmanager:GetSecretValue"
+      ],
+      "Effect": "Allow",
+      "Resource": "arn:aws:secretsmanager:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:secret:datadog_api_key*"
+    }
+  ]
+}
+EOF
+}
+
+data "aws_caller_identity" "current" {}
--- a/terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
+++ b/terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -x
+exec > >(tee /tmp/user-data.log|logger -t user-data ) 2>&1
+
+#SETTING UP HOSTNAME
+
+REGION="${region}"
+IP=`curl -s http://169.254.169.254/latest/meta-data/public-ipv4`
+echo "aws-$REGION-$IP" > /etc/hostname
+hostname "aws-$REGION-$IP"
+
+#SETUP DATADOG
+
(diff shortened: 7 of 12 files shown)


terraform/modules/cloud/aws/deploy/fleet/iam.tf
```diff
@@ -0,0 +1,43 @@
...
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "kms:Decrypt"
+      ],
+      "Effect": "Allow",
+      "Resource": "*"
```
dincho: Can we limit the instances that are allowed to decrypt? E.g. only epoch instances

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -0,0 +1,41 @@
...
+#SETTING UP HOSTNAME
+
+REGION="${region}"
+IP=`curl -s http://169.254.169.254/latest/meta-data/public-ipv4`
+echo "aws-$REGION-$IP" > /etc/hostname
+hostname "aws-$REGION-$IP"
+
+#SETUP DATADOG
+apt-get update
+apt-get install awscli -y
```
dincho: I prefer to have to in the `package` image.

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -0,0 +1,41 @@
...
+
+VERSION="0.16.0"
+
+FILE="epoch-$VERSION-ubuntu-x86_64.tar.gz"
+URL="https://github.com/aeternity/epoch/releases/download/v$VERSION/$FILE"
+
+wget -q $URL -O /home/epoch/$FILE
+
+mkdir /home/epoch/node
+tar -xvf /home/epoch/$FILE -C /home/epoch/node
```
dincho: `-v` is obsolete verbosity

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -0,0 +1,41 @@
...
+chown -R epoch:epoch /home/epoch/node
+#START EPOCH & DATADOG
+sudo su -c "/home/epoch/node/bin/epoch start" epoch
+sudo service datadog-agent start
```
dincho: Could you please move `datadog start` to it's related section above?

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -0,0 +1,41 @@
...
+DATADOG_API_KEY=`aws --region eu-west-1 kms decrypt --ciphertext-blob fileb:///tmp/datadog_api_key_encrypted --output text --query Plaintext | base64 --decode`
+
+sed -i -- "s/DATADOG_API_KEY/$DATADOG_API_KEY/g" /etc/datadog-agent/datadog.yaml
+
+sed -i -- "s/region:unknown/region:${region}/g" /etc/datadog-agent/datadog.yaml
+sed -i -- "s/color:unknown/color:${color}/g" /etc/datadog-agent/datadog.yaml
+sed -i -- "s/env:unknown/env:${env}/g" /etc/datadog-agent/datadog.yaml
+#INSTALL EPOCH
+
+VERSION="0.16.0"
```
dincho: we should be able to pass the `EPOCH_VERSION` as TF variable.

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -0,0 +1,41 @@
...
+REGION="${region}"
+IP=`curl -s http://169.254.169.254/latest/meta-data/public-ipv4`
+echo "aws-$REGION-$IP" > /etc/hostname
+hostname "aws-$REGION-$IP"
+
+#SETUP DATADOG
+apt-get update
+apt-get install awscli -y
+
+DATADOG_API_KEY_ENCRYPTED="${datadog_api_key}"
```
dincho: I thought we'd fetch this secret from KMS in this script not passing as TF variable

terraform/modules/cloud/aws/deploy/fleet/iam.tf
acidsennui: iam profile is attached just to epoch instaces

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: it's encrypted string, that instance is decrypting on boot

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: sure we can

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: sure

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
dincho: yes, but I think that the instance should fetch the encrypted script too, not TF.

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: moved

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: removed

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: done

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: done

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: done with usage of AWS Secrets Manager. Only epoch spot nodes can read this secrets permissions granted by IAM profile attache do node.

Makefile
```diff
@@ -22,7 +22,7 @@ images: ansible/roles
 setup-infrastructure: ansible/roles check-deploy-env
 	cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python"' \
 		--tags "$(DEPLOY_ENV)" environments.yml
-	cd terraform && terraform init && terraform apply --auto-approve
+	cd terraform && terraform init && terraform apply -var epoch_version=0.16.0 --auto-approve
```
dincho: so far so good, let's make this Makefile variable with default to 0.16.0:

- `EPOCH_VERSION =? 0.16.0` at makefile top
- `-var epoch_version=$(EPOCH_VERSION)` on this and following lines

acidsennui: fixed

Makefile
```diff
@@ -1,3 +1,4 @@
+EPOCH_VERSION =? 0.16.0
```
lucafavatella: Why defaulting?

lucafavatella: I mean - silent unexpected default seems worse than error.

dincho:
> What is use case? Only initial provisioning of new node in an environment (e.g. not update of OS or epoch in existing node)?
> 
> If yes, then is that not already done via Packer, as documented, and with automatic build-and-upload by CI server by pushing to packer branch?
> 
> If terraform - introduced in #63 - is alternative to packer, then maybe a couple of lines are needed in readme (that in this way it would clarify how tf fits with packer, and whether terraform can be used both locally and via CI)?
> 
> (This is initial batch of comments - after I read answers I will need to understand what this PR changes compared to master - as tf is already in master.)

The main change of this PR is https://github.com/aeternity/infrastructure/pull/64/files#diff-d3b479e0cb63584581a60025e37dfa0a which does 2 things:
- finalize an instance setup during the boot by setting a hostname and any secrets (datadog only ATM)
- deploy the specified epoch version on the instance

As @acidsennui mentioned this is mainly needed for spot instances where AWS may dynamically destroy and create them.

The hearth of the instance setup is Ansible. We keep all the configuration in it's playbooks + all required secrets (datadog only ATM). Those playbooks can be run manually and by CI to setup newly created instances - one by one.

We started to build "instance images" both for OS and AWS (so called AMIs) to speedup the environments setup. This basically reverses the process of provisioning - we provision only one instance (the AMI and OS Image) with the same playbooks and then use that image to create new environments.
Still, I don't want to keep any secrets in the prebuild image, because at some point we could make it public so that anyone can use it.

Thus, this PR introduce setup finalization with secrets and auto deploying the epoch. Before this PR, once the environment is created, we need to wait the CI to run setup cronjobs or setup it manually (ansible playbook run).

After merging this, the infrastructure configuration will be much more flexible, you can spin up a lot of new instances in almost no time.

Makefile
dincho: Rethinking this, it seems wrong to me to have one version for all environments, thus I believe that this constant should be hardcoded in the terraform main file for each module e.g. https://github.com/aeternity/infrastructure/pull/64/files#diff-b89d1f9fe4d022751a376ef53e46021dR42

As TF configuration should represent the current state of the infrastructure, also we must have the option to set different epoch version per environment.

.circleci/config.yml
```diff
@@ -62,6 +67,20 @@ jobs:
           name: Dry run environments playbooks
           command: .circleci/docker-run.sh make test-setup-environments
 
+  check-terraform:
```
dincho: Could you please use the same keys conventions in the file already with `_` instead of `-`.

.circleci/config.yml
```diff
@@ -62,6 +67,20 @@ jobs:
...
+    <<: *container_config
+    steps:
+      - checkout
+      - *restore_venv
+      - *activate_venv
+      - *install_pip
+      - run:
+          name: Configuration lint terrafomr
```
dincho: typo in `terrafomr`

.circleci/config.yml
```diff
@@ -62,6 +67,20 @@ jobs:
...
+          command: make lint-terraform
+      - run:
+          name: Dry run terrafrom plan
```
dincho: typo in `terrafrom`

.circleci/config.yml
dincho: Also please move all the steps of this job into `build_check` job. Don't forget to run the command with `.circleci/docker-run.sh make your_recipe`

.circleci/config.yml
```diff
@@ -132,6 +151,22 @@ workflows:
               ignore:
                 - packer
 
+      - check:
```
dincho: this has been merged to `build_check`

.circleci/config.yml
```diff
@@ -132,6 +151,22 @@ workflows:
...
+          requires:
+            - build
+          filters:
+            branches:
+              ignore:
+                - packer
+
+      - check-terraform:
```
dincho: this should be merged to `build_check`

Makefile
```diff
@@ -1,3 +1,5 @@
+SITE_PACKAGES := $(shell python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
```
dincho: This is not needed after the refactoring. Looks like rebase artefact to me.

Makefile
```diff
@@ -7,7 +9,10 @@ images:
 	packer build packer/epoch.json
 
 setup-infrastructure: check-deploy-env
-	cd ansible && ansible-playbook --tags "$(DEPLOY_ENV)" environments.yml
```
dincho: rebase artefact

Makefile
```diff
@@ -54,16 +59,23 @@ test-openstack:
 		-t openstack/ae-environment.yml --enable-rollback --wait --dry-run
 
 test-setup-environments:
-	cd ansible && ansible-playbook --check -i localhost, environments.yml
-	cd terraform && terraform init && terraform plan
+	cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python3"' \
```
dincho: rebase artefact

Makefile
```diff
@@ -54,16 +59,23 @@ test-openstack:
...
+		--check -i localhost, environments.yml
+
 
+test-setup-environments-terraform: pip
```
dincho: you don't need the `pip` dependancy anymore

Makefile
```diff
@@ -7,7 +9,10 @@ images:
 	packer build packer/epoch.json
 
 setup-infrastructure: check-deploy-env
-	cd ansible && ansible-playbook --tags "$(DEPLOY_ENV)" environments.yml
+	cd ansible && ansible-playbook -e 'ansible_python_interpreter="/usr/bin/env python3"' \
+		--tags "$(DEPLOY_ENV)" environments.yml
+
+setup-infrastructure-aws: ansible/roles check-deploy-env
```
dincho: you don't need both dependancies anymore

.circleci/config.yml
```diff
@@ -188,7 +206,10 @@ workflows:
               only:
                 - master
     jobs:
-      - setup_uat
+      - setup_terraform
+      - setup_uat:
+          requires:
+            - setup-terraform
```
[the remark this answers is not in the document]
dincho: > why are not the setup jobs of the other environments updated as well e.g. setup_integration?

Terraform setups only UAT nodes so far

terraform/main.tf
```diff
@@ -69,6 +83,13 @@ module "aws_deploy-us-west-2" {
   spot_price    = "0.125"
   instance_type = "m4.large"
 
+  epoch = {
+    version = "0.18.0"
```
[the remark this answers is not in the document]
dincho: I like the flexibility to update an environment region by region.

As I explained above in https://github.com/aeternity/infrastructure/pull/64#issuecomment-407058487 the configuration *must* be hardcoded and represent the current state.

Having this in external variable will be much more error prone as multiple users (devops/ci) might have set different variables.

terraform/modules/cloud/aws/deploy/fleet/variables.tf
```diff
@@ -15,3 +15,8 @@ variable "vpc_id" {}
 variable "subnets" {
   type = "list"
 }
+
+variable "epoch" {
+  type    = "map"
+  default = {}
```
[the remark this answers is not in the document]
dincho: Obstacle from test configuration

dincho:
> An update/integration of README.md would be helpful - terraform seems not mentioned at all.

@lucafavatella see 1f8e360

ansible/setup.yml
```diff
@@ -34,9 +34,9 @@
         - project-user
 
     # Automatic cleanup is not supported. You must remove old keys manually.
-    - name: "Setup root SSH authorized keys"
+    - name: "Setup master SSH authorized keys"
       authorized_key:
-        user: "{{ ansible_user }}"
+        user: master
```
lucafavatella: I understand CI runs this script every day on existing environments. What is impact?

packer/epoch.json
```diff
@@ -52,7 +52,7 @@
     },
     {
       "type": "ansible",
-      "user": "master",
+      "user": "ubuntu",
```
lucafavatella: I notice this ubuntu is different from the master set in other script - unclear to me full picture. I guess this is something specific to packer and the ami used?

ansible/setup.yml
dincho: Should have no impact for CI runs. This is fix for packer builds where the playbook is run with ‘ubuntu’ user for AWS builds

packer/epoch.json
dincho: The difference is because AWS and OS are using different base images with different sudo users

dincho: We unify this during the image builds

PR #67, opened 2018-07-09
Deploy playbook

dincho:
in scope of https://www.pivotaltracker.com/story/show/158550692

The playbook is not used yet, but it's part of the migration. It's basically copy/paste from epoch repo with some file name changes, peer key files with changed ansible vault password and [db path reset](https://github.com/aeternity/infrastructure/compare/158550692_deploy_playbook?expand=1#diff-0a05600d18acbea9b59fd306c7e25899R61)

Deploy example: https://circleci.com/gh/aeternity/infrastructure/1254


ansible/deploy.yml
```diff
@@ -0,0 +1,255 @@
...
+      chain:
+        persist: true
+        db_path: "./db"
+      keypair:
+        dir: "keys"
+        password: "{{ vault_keys_password|default('secret') }}"
+      logging:
+        level: "{{ log_level|default('error') }}"
+      mining:
+        expected_mine_rate: "{{ expected_mine_rate|default(300000) }}"
```
lucafavatella: https://github.com/aeternity/epoch/pull/1323/files#diff-2ad1947fe9b869ffd23a6a86872e65e0R68 ?

ansible/templates/epoch.yaml
```diff
@@ -33,6 +33,7 @@ websocket:
 
 mining:
     autostart: true
+    beneficiary: {{ config.mining.beneficiary }}
```
lucafavatella: It looks to me this needs setting - see my other comment.

Makefile
```diff
@@ -1,11 +1,12 @@
 .DEFAULT_GOAL := lint
+DEPLOY_DOWNTIME ?= 0
+DEPLOY_DB_VERSION ?= ""
```
lucafavatella: I would expect no default here. What is usage of default? (Caller not setting DB dir is a bug in the caller for me.)

lucafavatella:
I understand corresponding de-duplication in epoch repo tracked in https://www.pivotaltracker.com/story/show/158550692

dincho:
> I understand corresponding de-duplication in epoch repo tracked in https://www.pivotaltracker.com/story/show/158550692

Yes, it's the "run X playbook ..." tasks

ansible/templates/epoch.yaml
```diff
@@ -0,0 +1,54 @@
+---
```
lucafavatella: I see this the same as epoch 4f3b0f7c - ok

Makefile
dincho: Fair point, removed in 839db7b

ansible/deploy.yml
```diff
@@ -0,0 +1,257 @@
...
+
+  vars:
+    project_root: "{{ ansible_env.HOME }}/node"
+    genesis_accounts_path: "{{ project_root }}/data/aecore/.genesis/accounts.json"
+    is_remote_package: false
+    public_ip: "{{ ansible_ssh_host|default(ansible_host)|default(inventory_hostname) }}"
+    local_peer_keys_dir: "files/peer_keys/{{ public_ip }}"
+    datadog_api_url: https://api.datadoghq.com/api/v1/downtime?api_key={{ vault_datadog_api_key }}&application_key={{ vault_datadog_app_key }}
+    downtime: "0"
+    db_version: ""
```
lucafavatella: ?

dincho: I amended previous commit so 0249283

Makefile
dincho: Amended to 0249283

ansible/group_vars/tag_env_integration/genesis_accounts.yml
```diff
@@ -0,0 +1,2 @@
+genesis_accounts:
```
lucafavatella: Is this file name meaningful? I saw changed from accounts.yml in epoch repo, though it looks like name does not matter so not reference to it needs update.

ansible/mnesia_backup.yml
```diff
@@ -0,0 +1,38 @@
+---
```
lucafavatella: Clarify whether this part.

ansible/group_vars/tag_env_integration/genesis_accounts.yml
dincho: No, the filename does not matter from execution point of view. I renamed it to have more meaningful name in this repo context/scope.

dincho:
> Just checking - tag_seed_uat inventory group not present?

I didn't migrated it as it's not used by automation, however it should be covered in https://www.pivotaltracker.com/story/show/158872869

ansible/mnesia_backup.yml
dincho: The file was added by accident. Now removed.

PR #68, opened 2018-07-11
Introduce infrastructure docker image

dincho:
As it can be seen from the README it greatly simplifies to usage of all the tools. Moreover I plan to use this image as source in the `epoch` repo to run the deployments (once migrated to this repo).

The CircleCI runs the builds and checks on a machine config of current branch source code, then all the commands on master are run inside the freshly built container. In the end that image is pushed to docker hub as `latest` tag.

in scope of https://www.pivotaltracker.com/story/show/158550692


README.md
```diff
@@ -56,6 +42,21 @@ There is a tiny bridge vault file `vault-env` that bridges the `INFRASTRUCTURE_A
 export INFRASTRUCTURE_ANSIBLE_VAULT_PASSWORD="top secret"
 ```
 
+## Docker image
+
+A Docker image `aeternity/infrastructure` is build and published to DockerHub. To use the image one should configure all the required credentials as documented above and run the container:
+
+```bash
+docker run -it --env-file env.list -v ~/.ssh/:/root/.ssh/ aeternity/infrastructure
```
dincho: Unfortunately sharing socket (ssh-agent) with docker does not work, thus this is the only working solution I can think of. Any better ideas are welcome.

ansible/inventory/aws.aws_ec2.yml
```diff
@@ -0,0 +1,9 @@
+# Ansible AWS EC2 dynamic inventory plugin
+# https://docs.ansible.com/ansible/2.6/plugins/inventory/aws_ec2.html
+plugin: aws_ec2
```
dincho: As this is plugin already, we don't need to carry the .py file below anymore.

ansible/vault-env
```diff
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/bin/sh
```
dincho: alpine does not ship with bash by default, and we don't need it anyway

ansible/inventory/aws.aws_ec2.yml
```diff
@@ -0,0 +1,9 @@
+# Ansible AWS EC2 dynamic inventory plugin
+# https://docs.ansible.com/ansible/2.6/plugins/inventory/aws_ec2.html
```
lucafavatella: I see 2.5, that is version pinned in requirements, has this too - https://docs.ansible.com/ansible/2.5/plugins/inventory/aws_ec2.html - so ok.

.circleci/config.yml
```diff
@@ -2,34 +2,24 @@ version: 2
 
 references:
   container_config: &container_config
-    setup_remote_docker:
-      docker_layer_caching: true
     docker:
-      - image: aeternity/builder
-        user: builder
+      - image: aeternity/infrastructure:master
```
lucafavatella: I admit I did not get my head around this PR fully though I would expect corresponding PR removing from builder Docker image [packer and terraform](https://github.com/aeternity/docker-builder/blob/ec4a8d4eb3c19b98319a16dd2d541f734c599f9a/Dockerfile#L32-L40), to be merged shortly after this?

lucafavatella: I see this PR is a chicken-and-egg one - introducing both the build+push of aeternity/infrastructure:master Docker image *and its usage*. So this can only work if this PR is merged well before the time of trigger of setup of environments (midday UTC ? every day so merge shall happen e.g. in afternoon or in very early morning - otherwise a setup env run would fail).

.circleci/config.yml
```diff
@@ -46,58 +36,43 @@ references:
...
 
   check:
-    <<: *container_config
+    <<: *machine_config
     steps:
       - checkout
-      - *restore_venv
-      - *activate_venv
-      - *install_pip
+      - *pull_current_image
```
lucafavatella: This seems to me another chicken-and-egg problem, and I do not see this self-healing for any branch (this check job is run for all branches but packer) unless a manual push of image for that branch is done.

@dincho Could you please elaborate?

lucafavatella: I see the image pulled here is used below in .circleci/docker-run.sh so this needs to be clarified.

Makefile
```diff
@@ -1,59 +1,43 @@
...
-
-ansible/roles: pip
-	cd ansible && ansible-galaxy install -r requirements.yml -p roles
-	touch ansible/roles
-
-images: ansible/roles
+images:
 	packer build packer/epoch.json
 
-setup-infrastructure: ansible/roles check-deploy-env
```
lucafavatella: I see [ansible/roles moved to Dockerfile](https://github.com/aeternity/infrastructure/pull/68/files#diff-3254677a7917c6c01f55212f86c57fbfR22) so the Makefile now assumes more to be run insider Docker container. Ok.

.circleci/config.yml
dincho: Yes.

ansible/inventory/aws.aws_ec2.yml
dincho: 2.5 is minimum requirement, the actual installed is 2.6

.circleci/config.yml
dincho: I'm not sure I understand your concern about timings, the cron auto-setup will run whatever i sin `aeternity/infrastructure:master` image. The configuration that it runs on master is kind of misleading, because it actually does not care of the current repo source code, but the latest `aeternity/infrastructure:master` which is in fact the latest master.

.circleci/config.yml
dincho: The image (per branch) is built in the `build` job above. The `check` job has `build` as requirement, thus `build` is always run before `check` and it's guaranteed that the image is already build.

However, the CircleCI machine scheduler may assign *the same* or different VM worker to the `check` job than the `build` job, in this case ([example workflow](https://circleci.com/workflow-run/d0e8ff39-08bf-4636-ab81-7bd54bc1e8ee)) the image is not seen by the `check` job and has to be pulled from a remote (dockerhub). 

In most of the cases it shall be already available on the `check` VM worker e.g. see the pull step in [this workflow](https://circleci.com/gh/aeternity/infrastructure/1186)

lucafavatella: > The image (per branch) is built in the build job above. The check job has build as requirement

ah - thanks, I had missed this.

Other issue I see is that concurrent jobs running for same branch may overlap hence the check job may pull the image built for another workflow on the same branch.

I see no job requiring build job rather than check. Any reasons for not merging the build and check jobs as one job - so to avoid the potential overlap of jobs for same branch (and reduce complexity)?

dincho: Good idea, see 1deb63d

PR #75, opened 2018-07-27
Fix packer openstack builds

dincho:
As our openstack provider updated their version we faced packer SSH keys issue https://github.com/hashicorp/packer/issues/2526
This installs openssl to allow the workaround in the upsteam fix.

This also exposes OS_TENANT_ID to docker environment as the OS_PROJECT_ID is not supported by packer.


dincho:
closing of flavour of #76 because of branch name issue with docker tags

PR #81, opened 2018-08-06
Fix default HTTP check URL

dincho:
The default URL is used when pre-building instance images.

relates to: https://www.pivotaltracker.com/story/show/159518269 and https://www.pivotaltracker.com/story/show/159491502

I didn't confirmed yet, but invalid URL in this checks cause datadog error, which in turn might prevent sending other metrics as well.

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -44,7 +44,7 @@
         init_config:
         instances:
           - name: API
-            url: "{{ api_base_uri|default('127.0.0.1') }}/top"
+            url: "{{ api_base_uri|default('http://127.0.0.1:3013/v2') }}/top"
             # Default is false, i.e. emit events instead of service checks.
             # Recommend to set to true.
             skip_event: true


ansible/monitoring.yml
```diff
@@ -44,7 +44,7 @@
         init_config:
         instances:
           - name: API
-            url: "{{ api_base_uri|default('127.0.0.1') }}/top"
+            url: "{{ api_base_uri|default('http://127.0.0.1:3013/v2') }}/top"
```
lucafavatella: Could default for `api_base_uri` be removed from here?

dincho: Not sure what you mean exactly, could you please elaborate with more details ?

dincho: If you mean to move the default to the api_base_uri definition in `ansible/group_vars/tag_role_epoch/vars.yml`, it would be rather complex task because of the vaults. Right now we intentionally don't tag the `image.yml` playbook run with `tag_role_epoch`

To be able to do that we should first take out the secrets in `ansible/group_vars/tag_role_epoch/vault.yml` to the environment somehow.

PR #82, opened 2018-08-07
Allow outgoing UDP:123 to anywhere (NTP)

dincho:
https://www.pivotaltracker.com/story/show/158864551

### DIFF
--- a/openstack/ae-environment.yml
+++ b/openstack/ae-environment.yml
@@ -102,6 +102,7 @@ resources:
           port_range_max: { get_param: external_sync_port }
         - { direction: ingress, remote_ip_prefix: 0.0.0.0/0, protocol: tcp, port_range_min: 22, port_range_max: 22}
         - { direction: ingress, remote_ip_prefix: 0.0.0.0/0, protocol: icmp }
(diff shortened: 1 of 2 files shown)


openstack/ae-environment.yml
```diff
@@ -102,6 +102,7 @@ resources:
           port_range_max: { get_param: external_sync_port }
         - { direction: ingress, remote_ip_prefix: 0.0.0.0/0, protocol: tcp, port_range_min: 22, port_range_max: 22}
         - { direction: ingress, remote_ip_prefix: 0.0.0.0/0, protocol: icmp }
+        - { direction: egress, remote_ip_prefix: 0.0.0.0/0, protocol: udp, port_range_min: 123, port_range_max: 123}
```
lucafavatella: I expect this is for request to server or ack to server, though I would expect ingress too for NTP response from server?

dincho: AFAIK regardless that UDP is stateless the firewall still tracks the state, but deletes it with idle timer. Otherwise how would you know the client port to allow in ingress :)

PR #85, opened 2018-08-10
Add seed peers check

dincho:
- run after each UAT setup
- run hourly

https://www.pivotaltracker.com/story/show/159498903

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,13 @@ lint:
 	packer validate packer/epoch.json
 	cd terraform && terraform init && terraform validate && terraform fmt -check=true -diff=true
 
+# Keep in sync from https://github.com/aeternity/epoch/blob/master/config/sys.config
+check-seed-peers:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -70,6 +70,13 @@ lint:
 	packer validate packer/epoch.json
 	cd terraform && terraform init && terraform validate && terraform fmt -check=true -diff=true
 
+# Keep in sync from https://github.com/aeternity/epoch/blob/master/config/sys.config
+check-seed-peers:
+	curl -fs -m 5 http://52.10.46.160:3013/v2/peer/key | grep -q 'QU9CvhAQH56a2kA15tCnWPRJ2srMJW8ZmfbbFTAy7eG4o16Bf'
```
lucafavatella: I see same as https://github.com/aeternity/epoch/blob/34f0b50069b76f448551b8a9902b835d718eb084/config/sys.config#L31-L34

PR #86, opened 2018-08-10
Disable serial/rolling deployments

dincho:
https://www.pivotaltracker.com/story/show/159500501

I don't think we need rolling/serial deployments as of now after we introduced two phase manual UAT deployments.

Disabling the serial deployments should speedup the total deployment time a bit.

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -27,7 +27,6 @@
   # Facts are already collected in the previous playbook
   gather_facts: no
   max_fail_percentage: 25
-  serial: "25%"
 
   vars:
     project_root: "{{ ansible_env.HOME }}/node"


ansible/deploy.yml
```diff
@@ -27,7 +27,6 @@
   # Facts are already collected in the previous playbook
   gather_facts: no
   max_fail_percentage: 25
```
lucafavatella: As no other item to run after this, this line is now unused?

dincho: It still should fail and stop if 25% of the batch fails. We just removed the batch configuration, now the batch = all the hosts

dincho:
> Please update https://github.com/aeternity/epoch/wiki/Release-process

https://github.com/aeternity/epoch/wiki/Release-process/_compare/249d9bf6f2a58ccdedf7758f275b957693ccae89...2fc89e20aa9b36bae8d6085086e4010bfaa74988

PR #88, opened 2018-08-13
Terraform install latest 

acidsennui:
pt: https://www.pivotaltracker.com/story/show/159699819

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -39,7 +39,7 @@ module "aws_deploy-ap-southeast-1" {
 
   epoch = {
     version = "0.18.0"
-    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-latest-ubuntu-x86_64.tar"
+    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-latest-ubuntu-x86_64.tar.gz"
 
     # The string must be bash escaped
     beneficiary = "ak\\$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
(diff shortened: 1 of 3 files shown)


terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -26,15 +26,12 @@ sudo service datadog-agent start
...
 #INSTALL EPOCH
 
-VERSION="${epoch_version}"
+URL="${epoch_package}"
 
-FILE="epoch-$VERSION-ubuntu-x86_64.tar.gz"
-URL="https://github.com/aeternity/epoch/releases/download/v$VERSION/$FILE"
-
-wget -q $URL -O /home/epoch/$FILE
+wget -q $URL -O /home/epoch/epoch.tar.gz
```
lucafavatella: Above, value of variable `URL` is assigned with quotes `"` while here no quotes used: shall quotes be here too?

I see name is inject from other config file, is URL, and potentially may contain strange characters. Though it does not matter for now so looks ok.

terraform/main.tf
```diff
@@ -38,7 +38,7 @@ module "aws_deploy-ap-southeast-1" {
   instance_type = "m4.large"
 
   epoch = {
-    version = "0.19.0"
+    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-latest-ubuntu-x86_64.tar.gz"
```
lucafavatella: How does guarantee comes that `s3.eu-central-1.amazonaws.com` serves latest version of file? (I mean - eu-central-1 may have stale copy while another asia or so have a recent one.)

lucafavatella: Does this terraform integration print the hash of downloaded artefact (minimum)?

michalbagrowski: we out our images only in eu-central-1, aws it self to not synch it cross region

michalbagrowski: terrafrom do not download this file, it's isntance that do this on boot, can add this there

acidsennui: we upload image just to eu-central-1 s3 bucket, aws it self do not synch it to other regions

acidsennui: terraform it self do not touch this file, it's used with `user_script` in boot of instance, i can add there hash if required

lucafavatella: Not required. (I see it is release bucket so not many re-upload.)

dincho:
> This seems to depend on aeternity/epoch#1437 so cannot be merged now?

Actually it does not until next release, because the `latest` artefact is manually put in place already.

PR #97, opened 2018-09-11
Replace '$' with '_' in addresses

hanssv:
Solves https://www.pivotaltracker.com/story/show/160254986


hanssv:
https://github.com/aeternity/epoch/pull/1544

PR #102, opened 2018-09-11
Use fixed infrastrcuture repo version during boot

dincho:


terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -6,7 +6,7 @@ exec > >(tee /tmp/user-data.log|logger -t user-data ) 2>&1
 
 export INFRASTRUCTURE_ANSIBLE_VAULT_PASSWORD=`aws --region ${region} secretsmanager get-secret-value --secret-id ansible_vault_password --output text --query 'SecretString'`
 
-git clone -b master --single-branch https://github.com/aeternity/infrastructure.git /infrastructure
+git clone -b v1.0 --single-branch https://github.com/aeternity/infrastructure.git /infrastructure
```
lucafavatella: I see v1.0 as 35c05cb i.e. without _ replacing $

lucafavatella:
I see from Circle CI config file that terraform related to UAT only.

dincho:
> Is 044e1a9 needed for UAT?

No. It's tricky.

That image is already built and published. The infrastructure configuration still used the master branch (where it's included), thus UAT will use that image.

However, the configuration changes the source code of the infrastructure that would be used to boot (initially setup+deploy) new instances.

PR #103, opened 2018-09-12
Remove datadog explicit hostname configuration

dincho:
Explicit hostname configuration confuses datadog very often because it's
not unique.
Removing explicit configuration will cause datadog autodiscovery to kick
in
https://docs.datadoghq.com/agent/faq/how-datadog-agent-determines-the-hostname/
that's the instance ID usually.

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -49,7 +49,6 @@
       process_config:
         enabled: "true" # has to be set as a string
       logs_enabled: true
-      hostname: "{{ ansible_fqdn }}"
       tags:
         - "platform:{{ platform|default('os') }}"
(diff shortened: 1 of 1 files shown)


dincho:
> If I understand that graph correctly `/bin/hostname -f` will be used?

No, the instance-id, because in our case it's EC2 default.

> agent-hostname: If a host name is explicitly set in the Agent configuration file.
hostname: If the DNS host name *is not an EC2 default* (e.g. ip-192-0-0-1).
instance-id: If the Agent can reach the EC2 metadata endpoint from the host.
hostname: Fall back on the DNS host name even if it is an EC2 default.

PR #106, opened 2018-09-13
Add datadog kind tag

dincho:
https://www.pivotaltracker.com/story/show/159620185

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -30,6 +30,7 @@
       set_fact:
         env: "{{ instance_tags['tags']['env'] | default('unknown') }}"
         color: "{{ instance_tags['tags']['color'] | default('unknown') }}"
+        kind: "{{ instance_tags['tags']['kind'] | default('peer') }}"
       when: ansible_ec2_instance_id is defined
 
 - name: Configure monitoring services (DataDog)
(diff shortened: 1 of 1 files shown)


ansible/monitoring.yml
```diff
@@ -30,6 +30,7 @@
       set_fact:
         env: "{{ instance_tags['tags']['env'] | default('unknown') }}"
         color: "{{ instance_tags['tags']['color'] | default('unknown') }}"
+        kind: "{{ instance_tags['tags']['kind'] | default('peer') }}"
```
lucafavatella: Why not explicitly tagging non-static AWS instances as `peer` (as done for `seed` [here](https://github.com/aeternity/infrastructure/pull/84/files#diff-b1a76c7f484f0748d6486ec8e6b80adaR40)), and placing default `unknown` here? Anyway debatable so ok.

ansible/monitoring.yml
```diff
@@ -55,6 +56,7 @@
         - "region:{{ region|default('eu-east') }}"
         - "env:{{ env|default('unknown') }}"
         - "color:{{ color|default('unknown') }}"
+        - "kind:{{ kind|default('peer') }}"
```
lucafavatella: I see in combination with already-merged https://github.com/aeternity/infrastructure/pull/84/files this delivers https://www.pivotaltracker.com/n/projects/2152594/stories/159620185

ansible/monitoring.yml
dincho: We still need a default to run the playbook on bare-metal for example

PR #108, opened 2018-09-21
Send system logs to datadog

dincho:
https://www.pivotaltracker.com/story/show/160280390

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -138,7 +138,7 @@
   handlers:
     - name: Restart rsyslogd
       service:
-        name: httpd
+        name: rsyslog
         state: restarted
       listen:
         - restart rsyslogd
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -44,10 +44,14 @@
     datadog_api_key: "{{ vault_datadog_api_key|default('DATADOG_API_KEY') }}"
     datadog_agent6: true
     datadog_config:
+      log_level: warning
+      log_to_console: false
       apm_enabled: false
       use_dogstatsd: true
       process_config:
         enabled: "true" # has to be set as a string
+      apm_config:
+        enabled: false
       logs_enabled: true
       hostname: "{{ ansible_fqdn }}"
       tags:
@@ -88,6 +92,10 @@
         init_config:
         instances:
         logs:
+          - type: tcp
+            port: 10518
+            service: epoch
+            source: syslog
           - type: file
             path: "/home/{{ project_user }}/node/log/epoch.log"
             service: epoch
@@ -118,3 +126,20 @@
 
   roles:
     - { role: Datadog.datadog }
+
+  post_tasks:
+    - name: Copy datadog.conf to rsyslog
+      copy:
+        src: files/rsyslog/datadog.conf
+        dest: /etc/rsyslog.d/datadog.conf
+        mode: 0600
+      notify: restart rsyslogd
+
+  handlers:
+    - name: Restart rsyslogd
+      service:
+        name: httpd
+        state: restarted
+      listen:
+        - restart rsyslogd
+
--- a/ansible/files/rsyslog/datadog.conf
+++ b/ansible/files/rsyslog/datadog.conf
@@ -1,0 +1,5 @@
+$template DatadogFormat,"%msg%\n"
+
+*.*  action(type="omfwd" target="localhost" port="10518" protocol="tcp" template="DatadogFormat"
+            action.resumeRetryCount="100"
+            queue.type="linkedList" queue.size="10000")


ansible/monitoring.yml
```diff
@@ -118,3 +126,20 @@
 
   roles:
     - { role: Datadog.datadog }
+
+  post_tasks:
+    - name: Copy datadog.conf to rsyslog
+      copy:
+        src: files/rsyslog/datadog.conf
+        dest: /etc/rsyslog.d/datadog.conf
```
lucafavatella: Who guarantees that rsyslog is actually installed?

I see OS images are:

https://github.com/aeternity/infrastructure/blob/b1bb61176a5457970162fba36656240206aeea30/ansible/environments.yml#L60

https://github.com/aeternity/infrastructure/blob/b1bb61176a5457970162fba36656240206aeea30/ansible/environments.yml#L137

that seem to be "ubuntu 16.04.3 with master user + python added".

ansible/monitoring.yml
```diff
@@ -118,3 +126,20 @@
...
+        mode: 0600
+      notify: restart rsyslogd
+
+  handlers:
+    - name: Restart rsyslogd
+      service:
+        name: httpd
```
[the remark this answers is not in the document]
lucafavatella: @dincho Could you please comment?

ansible/monitoring.yml
```diff
@@ -44,10 +44,14 @@
     datadog_api_key: "{{ vault_datadog_api_key|default('DATADOG_API_KEY') }}"
     datadog_agent6: true
     datadog_config:
```
lucafavatella: I see from ansible/requirements.yml that datadog plugin is https://galaxy.ansible.com/datadog/datadog v 2.1.0
so this var is identified [here](https://github.com/DataDog/ansible-datadog/blame/2.1.0/README.md#L28).

lucafavatella: Not clears to me who installs the datadog agent for processing this config.

ansible/monitoring.yml
dincho: Ubuntu versions that we use have rsyslog pre-installed.

dincho: While `infrastructure/ansible/environments.yml` is still in use, it's deprecated and I don't plan to support it. We're migrating to AWS/GCP soon, and it well be deleted in https://www.pivotaltracker.com/story/show/160761831

ansible/monitoring.yml
dincho: The datadog role here: https://github.com/aeternity/infrastructure/pull/108/files/be88057b106813d5e2754fc847f2a52ea84a16a1#diff-f1b92d458d2435cc2d62490c42d86397R128

ansible/monitoring.yml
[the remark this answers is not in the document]
dincho: Good catch, I run the playbook, but on preconfigured instance thus I didn't catch that.

ansible/monitoring.yml
lucafavatella: [`datadog` pip](https://pypi.org/project/datadog/) is installed [here](https://github.com/aeternity/infrastructure/blob/1dbc21209d65bc439330b0cb28eedca3b4d16aa2/ansible/setup.yml#L110).

ansible/monitoring.yml
```diff
@@ -44,10 +44,14 @@
...
+      log_level: warning
+      log_to_console: false
       apm_enabled: false
       use_dogstatsd: true
       process_config:
         enabled: "true" # has to be set as a string
+      apm_config:
+        enabled: false
```
lucafavatella: I see APM is this tracing https://docs.datadoghq.com/tracing/

dincho: Yes, great service, but without erlang support unfortunately.

ansible/monitoring.yml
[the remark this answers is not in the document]
dincho: Fixed in 2a36077

ansible/monitoring.yml
dincho: If you meant the role itself, it's installed with this file: https://github.com/aeternity/infrastructure/blob/master/ansible/requirements.yml#L1
by https://github.com/aeternity/infrastructure/blob/master/terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash#L13

ansible/monitoring.yml
```diff
@@ -44,10 +44,14 @@
...
       logs_enabled: true
```
lucafavatella: I see this documented [here](https://docs.datadoghq.com/logs/log_collection/?tab=tailexistingfiles).

Ok.

ansible/monitoring.yml
```diff
@@ -44,10 +44,14 @@
     datadog_api_key: "{{ vault_datadog_api_key|default('DATADOG_API_KEY') }}"
     datadog_agent6: true
     datadog_config:
+      log_level: warning
```
lucafavatella: I do not easily follow where this is documented or what it means. Does it mean to report severity warning or higher - implying that any input shall be structurally known to have a certain severity? Or does it plainly **add** warning to any message? (Even if informational.)

ansible/monitoring.yml
```diff
@@ -88,6 +92,10 @@
         init_config:
         instances:
         logs:
+          - type: tcp
+            port: 10518
+            service: epoch
```
lucafavatella: I understand this [`service`](https://docs.datadoghq.com/logs/log_collection/?tab=tailexistingfiles) is just a label that datadog **adds** to the data it reads from port. I.e. it is not a filter on which lines of the syslog are meant to be sent. And has no connection whatsoever with Unix process names (or with which command a process was launched).

That leads to the question: does this mean that anything that goes in syslog is sent to datadog? (I see no filtering in ansible/files/rsyslog/datadog.conf )

And if yes, all with tag `epoch`? (Not sure if this will cause confusion in the UI - as if logs were from app.)

And is the severity filtered at all? I see nothing in "ansible/files/rsyslog/datadog.conf" and I do not understand `log_level`.

ansible/monitoring.yml
dincho: This is datadog configuration, it lowers (datadog) logging severity to warning from info (the default). This is to reduce useless logs that we start feeding to the logs collector (still datadog in our case).

ansible/monitoring.yml
dincho: Yes, `service: epoch` is for filtering purposes (in the UI). We have other services too (not yet feeding datadog logs). This is further narrowed by `source: syslog` to distinguish from epoch daemon logs itself. I suppose we can choose better name for the service in future, this was already used.

Anything that goes to syslog is send to the logs collector, we might narrow the scope once we have more statistics in few days. My smoke tests doesn't show a lot of records.

Yes, all the records are "tagged" as `service: epoch` and `source: syslog`

The `log_level` above is not related to the syslog configuration, it only lower the datadog agent severity.

PR #109, opened 2018-09-25
Add AWS London region

dincho:
in scope of https://www.pivotaltracker.com/story/show/160448926


dincho:
> ansible/environments.yml ?

I'll remove the OS setup once we have dual AWS running

dincho:
> packer/cleanup-ami-and-snapshots.py ?

The region is already there: https://github.com/aeternity/infrastructure/blob/master/packer/cleanup-ami-and-snapshots.py#L9

PR #111, opened 2018-09-25
Add dev environments to AWS

dincho:
completes https://www.pivotaltracker.com/story/show/160448926

after https://github.com/aeternity/infrastructure/pull/109, https://github.com/aeternity/infrastructure/pull/110

Successful deploy: https://circleci.com/gh/aeternity/epoch/33125 (note that openstack is off due to https://github.com/aeternity/infrastructure/pull/112, thus only AWS nodes has been deployed).

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -19,6 +19,12 @@ provider "aws" {
 }
 
 provider "aws" {
+  version = "1.24"
+  region  = "eu-west-2"
+  alias   = "eu-west-2"
+}
+
+provider "aws" {
   version                 = "1.24"
   region                  = "us-west-2"
   alias                   = "us-west-2"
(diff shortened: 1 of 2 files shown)


terraform/main.tf
```diff
@@ -88,3 +94,39 @@ module "aws_deploy-us-west-2" {
     aws = "aws.us-west-2"
   }
 }
+
+module "aws_deploy-dev1-eu-west-2" {
```
lucafavatella: I see correspondent UAT section has no env in the name so may cause confusion.

lucafavatella:
If this is multistage solution of https://www.pivotaltracker.com/n/projects/2152594/stories/160448926 please identify steps, and we need to clarify timing interaction with testnet deployment (and associated development testing needs - so dev1/dev2 downtime may be relevant).

dincho:
@lucafavatella All the OS environments will be removed in scope of https://www.pivotaltracker.com/story/show/160761831 after this and the other related stories are merged.

I'll update datadog links once this is merged, because I don't know the IP addresses in advance.

In reagards to CircleCI, if you meant to remove [the setup of dev environments](https://github.com/aeternity/infrastructure/blob/master/.circleci/config.yml#L136-L152) I plan to do in https://www.pivotaltracker.com/story/show/160761831


To summarise: I plan to create couple more PR's that **add** integration and fast_integration environments to AWS. No dawn-time of the environments is expected, just growing the node count. After that we'll kill the OS environments (dev1, dev2, integration, fast_integration).

dincho:
@lucafavatella I fixed missing job dependancy 802b76e

terraform/main.tf
dincho: Yes, it may need separate manual migration, thus not included here.

PR #113, opened 2018-09-25
Configure peers auto-discovery for non-uat environments

dincho:
in scope of https://www.pivotaltracker.com/story/show/160448926


dincho:
This won't work during the boot because of Ansible SSH requirement to fetch each node peer key. It's not feasible to allow each node to ssh to all other nodes.

PR #119, opened 2018-10-02
Fix manage-node playbook epoch configuration

dincho:


Makefile
```diff
@@ -39,8 +39,12 @@ manage-node: check-deploy-env
 ifndef CMD
 	$(error CMD is undefined, supported commands: start, stop, restart, ping)
 endif
-	cd ansible && ansible-playbook --limit="tag_env_$(DEPLOY_ENV):&tag_role_epoch" \
-		--extra-vars="cmd=$(CMD)" manage-node.yml
+	cd ansible && ansible-playbook \
+		--limit="tag_env_$(DEPLOY_ENV):&tag_role_epoch" \
+		-e env=$(DEPLOY_ENV) \
+		-e db_version=0 \
```
lucafavatella: I see this make target is not reset-net so not touching DB. So ok.

lucafavatella: I notice that user_data.bash uses db_version=0. Does it mean that when initialized starts at folder 0, then at next deploy from epoch repo it uses actual dir? (It may surprise DevOps person - assuming that epoch config on all nodes of environment points to same db folder.)

PR #121, opened 2018-10-03
Add UAT module dependancies

dincho:
While originally the solution was planned to split the configuration and use separate commands to run the provisioning, I find this solution more robust regardless of the used TF workaround.

The dependancy is set per `color` so that at any given time at least one seed node per `color` should be available. That would guarantee at least 2 seed nodes for UAT.

Finishes: https://www.pivotaltracker.com/story/show/160690120

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -72,6 +72,8 @@ module "aws_deploy-eu-central-1" {
   providers = {
     aws = "aws.eu-central-1"
   }
+
+  depends_on = ["${module.aws_deploy-ap-southeast-1.static_node_ips}"]
 }
 
 module "aws_deploy-us-west-2" {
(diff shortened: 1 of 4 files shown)


terraform/main.tf
```diff
@@ -72,6 +72,8 @@ module "aws_deploy-eu-central-1" {
   providers = {
     aws = "aws.eu-central-1"
   }
+
+  depends_on = ["${module.aws_deploy-ap-southeast-1.static_node_ips}"]
```
lucafavatella: From terraform/main.tf I see blue as:
aws_deploy-ap-southeast-1
aws_deploy-eu-central-1

So this makes sense.

terraform/main.tf
```diff
@@ -114,6 +116,8 @@ module "aws_deploy-uat-eu-west-2" {
   providers = {
     aws = "aws.eu-west-2"
   }
+
+  depends_on = ["${module.aws_deploy-us-west-2.static_node_ips}"]
```
lucafavatella: From terraform/main.tf I see green as:
aws_deploy-us-west-2
aws_deploy-uat-eu-west-2

So this makes sense.

dincho:
> It would be great if you included this in the merge commit so the intention of the changes is clear in the repo history.

sure, see https://github.com/aeternity/infrastructure/commit/ffa71eb394586ec9be23cf14da91be97cac2a871

PR #123, opened 2018-10-04
Test ansible playbooks in CI

dincho:


.circleci/config.yml
```diff
@@ -83,6 +83,28 @@ jobs:
       - run:
```
lucafavatella: So pushed to Docker Hub regardless of CI. Though it was this way before too. Ok.

.circleci/config.yml
```diff
@@ -83,6 +83,28 @@ jobs:
...
+      - run:
+          name: Prepare Ansible test environment
+          command: |
+            # Cleanup ansible environment on "remote" host. Order is important.
+            sudo apt-get remove python-urllib3 python-setuptools
+            sudo apt-get update && sudo apt-get install -y python-dev python-pip
+            sudo /usr/bin/pip2 install -U setuptools
+            # Allow ansible runs from "controller" host
+            pip install -r requirements.txt
+            ansible-galaxy install -r ansible/requirements.yml
```
lucafavatella: Placing most instructions in Makefile would ease test locally - out of CI. Though ok.

.circleci/config.yml
```diff
@@ -83,6 +83,28 @@ jobs:
...
+      - run:
+          name: Test ansible setup playbook
+          command: |
+            .circleci/test-ansible-playbook.sh ansible/setup.yml
```
lucafavatella: Which environment(s) is this command running against?

.circleci/config.yml
```diff
@@ -83,6 +83,28 @@ jobs:
...
+      - run:
+          name: Test ansible monitoring playbook
+          command: |
+            .circleci/test-ansible-playbook.sh ansible/monitoring.yml
```
lucafavatella: Which environment(s) is this command running against?

.circleci/config.yml
```diff
@@ -83,6 +83,28 @@ jobs:
...
+      - run:
+          name: Test ansible deploy playbook
+          command: |
+            .circleci/test-ansible-playbook.sh ansible/setup.yml -e env=test
```
lucafavatella: deploy or setup?

.circleci/config.yml
dincho: I doubt anyone going to test this locally

.circleci/config.yml
dincho: We use the notion of the environments only for deployment purposes. Otherwise the scope of execution is controlled by ansible inventory

.circleci/config.yml
dincho: See the above comment ^^

.circleci/config.yml
dincho: good catch

dincho: see https://github.com/aeternity/infrastructure/pull/123/commits/43459e22352842b845e5dd7dbc3df1cbfeb5397e and https://github.com/aeternity/infrastructure/pull/123/commits/dbf01ce0e32362fb94ea980dc6cac5c35cfd8f4a

PR #124, opened 2018-10-11
Add missing pip dependancy (local aws runs)

dincho:
The monitoring playbook autodiscover AWS tags and needs this lib to call the AWS API.

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -114,6 +114,7 @@
         - awscli
         - ansible
         - boto
+        - boto3
         - datadog
       tags:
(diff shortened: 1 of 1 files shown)


ansible/setup.yml
```diff
@@ -114,6 +114,7 @@
         - awscli
         - ansible
         - boto
+        - boto3
```
lucafavatella: Is boto needed then, given that there is boto3?

dincho: yes, unfortunately different libs use different versions.

PR #126, opened 2018-10-19
Bump bootstrap version

dincho:
- bump UAT bootstrap version to latest supported
- stick integration boostrap version to allow BC incompatible changes to
to be merged to master

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -36,7 +36,7 @@ module "aws_deploy-ap-southeast-1" {
   source            = "modules/cloud/aws/deploy"
   env               = "uat"
   color             = "blue"
-  bootstrap_version = "v1.1"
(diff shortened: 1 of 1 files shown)


terraform/main.tf
```diff
@@ -165,7 +165,7 @@ module "aws_deploy-dev2-eu-west-2" {
 module "aws_deploy-integration-eu-west-2" {
   source            = "modules/cloud/aws/deploy"
   env               = "integration"
-  bootstrap_version = "master"
```
lucafavatella: I see dev1 and dev2 still master. Intended?

dincho: Yes, I'd like to have an option to test the master via dev environments, while I usually spin up new environment, I'd like to keep this option.

PR #127, opened 2018-10-19
Hashicorp Vault integration to Ansible playbooks and infrastructure CI

dincho:
closes https://www.pivotaltracker.com/story/show/160452874 and https://www.pivotaltracker.com/story/show/160452979

The setup has been tested based on https://github.com/aeternity/infrastructure/commit/ece99f7c2f57c0320fb013908fe09eafaf062d73 in [CircleCI](https://circleci.com/gh/aeternity/infrastructure/4058)

However, old environment variables are not yet removed from CircleCI to prevent blocking the workflow until PR merge.


.circleci/config.yml
```diff
@@ -95,21 +100,32 @@ jobs:
             ansible-galaxy install -r ansible/requirements.yml
       - run:
           name: Test ansible setup playbook
+          environment:
+            - ANSIBLE_CALLBACK_WHITELIST: profile_tasks
```
lucafavatella: I see [key](https://docs.ansible.com/ansible/2.6/reference_appendices/config.html#envvar-ANSIBLE_CALLBACK_WHITELIST) and [value](https://docs.ansible.com/ansible/2.6/plugins/callback/profile_tasks.html) documented. Ok.

lucafavatella: I see this setting enables logging of timing of Ansible tasks. I see already present at ansible/ansible.cfg though, as these test calls do not `cd` to `ansible/` before running, they do not pick that up. Ok.

ansible/vars/epoch/uat.yml
```diff
@@ -1,6 +1,11 @@
 api_base_uri: http://{{ public_ipv4 }}:{{ epoch_config.http.external.port }}/v2
-
 configure_peers: false
+seed_peers:
+  - "52.10.46.160"
+  - "18.195.109.60"
+  - "13.250.162.250"
+  - "18.130.148.7"
```
lucafavatella: I see same 4 as at https://github.com/aeternity/epoch/blob/f7a0d9d38c372c2b37b2d6a19fa6768ee4804e64/docs/release-notes/RELEASE-NOTES-0.25.0.md#inspect-the-testnet
Ok.

ansible/vars/epoch/default.yml
```diff
@@ -1,6 +1,8 @@
 api_base_uri: http://{{ public_ipv4 }}:{{ epoch_config.http.external.port }}/v2
 genesis_accounts:
   "ak_UAzhn9rAQg568v6Hwt3w2HPaQb9X9Nw6JbLmnv7trhmGmWGGp": 100000000001
+configure_peers: true
```
lucafavatella: Shall [`configure_peers|default(true)`](https://github.com/aeternity/infrastructure/blob/c2326292a5c04801136e7916c61f584bdab531c0/ansible/templates/epoch.yaml#L2) be modified to check that `configure_peers` is mandatory, though without default `true` or `false` value, so to prevent typos? (It seems always specified now.) Totally optional of course.

dincho: I would like to refactor this variable name in future, maybe I'll rethink it's defaults and it should be mandatory or not then

import-secrets.sh
```diff
@@ -23,4 +23,5 @@ if [ -n "$VAULT_TOKEN" ]; then
     export DOCKER_USER=$(echo $DOCKERHUB_CREDS | grep -o 'username [^ ]*' | awk '{print $2}')
     export DOCKER_PASS=$(echo $DOCKERHUB_CREDS | grep -o 'password [^ ]*' | awk '{print $2}')
     export DATADOG_API_KEY=$(vault read -field=api_key secret/datadog/deploy)
+    export ROCKET_HOOK_URL=$(vault read -field=core-dev secret/rocketchat/prod/hooks)
```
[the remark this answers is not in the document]
dincho: Yes, that was the case, but I changed it to prevent misleading in aaf2824

PR #131, opened 2018-10-24
Install Vault to node machine images

dincho:
Allow bootstraping using SCM

In scope of https://www.pivotaltracker.com/story/show/161142747


dincho:
> I see. Then, there is no other suitable solution that has many real advantages than the current. Can a symbolic link be used instead of copying?

Yes, but I would not link `/usr/bin` to `/usr/src`, at least my understanding for `/usr/src` is that an operator can delete its contents anytime to free up space.

Using `/usr/bin/vault-{{version}}` would harm the autocomplete. Not sure it's good idea to have subdirs in the `/usr/bin` either, so what location do you suggest to keep the versioned binary, so we can link `/usr/bin/vault` to one of them?

valkov-ivaylo:
> Yes, but I would not link `/usr/bin` to `/usr/src`, at least my understanding for `/usr/src` is that an operator can delete its contents anytime to free up space.

I agree.
 
> Using `/usr/bin/vault-{{version}}` would harm the autocomplete. Not sure it's good idea to have subdirs in the `/usr/bin` either, so what location do you suggest to keep the versioned binary, so we can link `/usr/bin/vault` to one of them?

No sub-directories are allowed in /usr/bin. [FSH 4.4.2 Requirements](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrbinMostUserCommands)

```
4.4.2. Requirements
There must be no subdirectories in /usr/bin.
```

A good choice might be /opt.


FSH:  [[1]](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#optAddonApplicationSoftwarePackages)

```
3.13.1. Purpose
/opt is reserved for the installation of add-on application software packages.

A package to be installed in /opt must locate its static files in a separate /opt/<package> or /opt/<provider> directory tree, ...
```
Debian/Ubuntu based:   [[2]](https://wiki.debian.org/FilesystemHierarchyStandard) 
```
/opt/

    Add-on application software packages
    Pre-compiled, non ".deb" binary distribution (tar'ed..) goes here.

        /opt/bin/ : Same as for top-level hierarchy 
```

It could stay as is, but I am still not completely sure how it will affect a machine long-term.

dincho:
> It could stay as is, but I am still not completely sure how it will affect a machine long-term.

Most of the instances are spot. `/opt` would good choice otherwise.
I could change it in few minutes, but I'd need more time to test it, not sure it's worth it.

PR #133, opened 2018-10-25
Remove password for signing keys.

fabiankrol:
Related epoch PR: https://github.com/aeternity/epoch/pull/1681

By default `keys.peer_password` was resuing `keys.password`. That's why i'm renaming key.

### DIFF
--- a/ansible/vars/epoch/default.yml
+++ b/ansible/vars/epoch/default.yml
@@ -16,7 +16,7 @@ epoch_config:
 
   keys:
     dir: keys
-    password: secret
+    peer_password: secret
 
   chain:
     persist: true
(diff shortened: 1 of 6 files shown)


dincho:
Is this BC?

fabiankrol:
@dincho while adding signing keys rotation we decided to remove associated disk files. 
`password` configuration was used to decrypt keys from file - there is no longer need for that. 
However we still using this workflow for peer keys. Default `peer_password` was set up as `password` but that's no longer valid: https://github.com/aeternity/epoch/pull/1681/files#diff-12fc846715f3180b6b49156b1e022d4eR524

dincho:
@fabiankrol What I meant is - is the `password` configuration option required in 0.24.0 ?

fabiankrol:
I misunderstood `BC` meaning, sorry 😂 

It's not required. `password` in optional with the default value set to `secret`

PR #134, opened 2018-10-25
Allow AWS credentials role configuration

dincho:
in scope of https://www.pivotaltracker.com/story/show/160452961

### DIFF
--- a/import-secrets.sh
+++ b/import-secrets.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+AWS_CREDS_ROLE="${AWS_CREDS_ROLE:-epoch-fleet-manager}"
+
 if [ -n "$VAULT_GITHUB_TOKEN" ]; then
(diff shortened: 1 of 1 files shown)


import-secrets.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+AWS_CREDS_ROLE="${AWS_CREDS_ROLE:-epoch-fleet-manager}"
```
lucafavatella: I see this used in CI config and in docker entrypoint. So I expect that this is needed not for CI config but for docker entrypoint when called by epoch repo?

Also I expect configurable name for keeping roles separate between the two repos?

dincho: > Also I expect configurable name for keeping roles separate between the two repos?

Exactly

> So I expect that this is needed not for CI config but for docker entrypoint when called by epoch repo?

Docker entrypoints are not run by default by CircleCI, we run it explicitly (CircleCI supports labels, but the explicit sounds good for me at least for now, I may change it in future).

I didn't fully got your question tho.

PR #135, opened 2018-10-26
Reliable AWS secrets wait handling

dincho:


import-secrets.sh
```diff
@@ -29,5 +29,8 @@ if [ -n "$VAULT_TOKEN" ]; then
 
     # AWS dynamic credentials are eventually consistent - add a delay
     # https://www.vaultproject.io/docs/secrets/aws/index.html#usage
-    sleep 10
+    # Retry each second with ~20s timeout
+    for i in {1..20}; do
```
lucafavatella: I assume that shall can be relied to be [Bash](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html).

dincho:
closing in favour of #137

PR #138, opened 2018-10-31
Get rid of IAM role creation with TF

dincho:
As the CI credentials has been swtiched to STS AWS, this strategy does
not allow IAM access, thus TF cannot create IAM in an automated way
anymore. The IAM epoch role is managed "outside" of this repo.


dincho:
> (Side note: Reduction of usage of acronyms in commit messages and PR description would ease review.)

Sorry about that, the only acronym that's not used in the official documentation of the tools is TF = Terraform.

PR #139, opened 2018-10-31
Fix Vault install tasks

dincho:


ansible/setup.yml
```diff
@@ -124,24 +124,25 @@
...
         remote_src: true
       tags:
         - vault-install
 
     - name: Install Vault binary
       copy:
-        src: /usr/src/{{ vault_version }}/vault
+        src: /usr/src/vault-{{ vault_version }}/vault
         dest: /usr/bin/vault
+        remote_src: true
```
dincho: The actual fix being this one, but I also noticed the above path issues.

This is not catch by the CI tests/runs because both controller and remote are the same machine -> the file exists in both machines.

PR #140, opened 2018-11-01
Tag Docker and upload images with CircleCI on git tag.

valkov-ivaylo:
Closes https://www.pivotaltracker.com/n/projects/2152594/stories/161507366


.circleci/config.yml
```diff
@@ -265,6 +277,15 @@ workflows:
             branches:
               only: master
 
+      - docker_push_tag:
+          requires:
+            - build_check
```
dincho: This dependancy should be either removed or the same  tags filter should be added to it.
Not sure what's the value of `CIRCLE_BRANCH` during a tagging in `build_check` job

PR #142, opened 2018-11-01
Add CSM SSH support

dincho:
This can be tested by following the authentication introductions in the README.md and setting `VAULT_ADDR` and `VAULT_GITHUB_TOKEN`, ping me if you don't know the `VAULT_ADDR`.

Then run the branch container:
```bash
docker run -it --env-file env.list aeternity/infrastructure:160452952_csm_ssh
```

The CA certification is installed only on a dev2 node with address `3.8.20.123`, so:

```
make ssh HOST=3.8.20.123
```


Makefile
```diff
@@ -78,7 +92,10 @@ ifndef DEPLOY_ENV
 	$(error DEPLOY_ENV is undefined)
 endif
 
+clean:
+	rm ~/.ssh/id_*
+
```
valkov-ivaylo: I find that a bit risky. It still depends where the Makefile is executed. If I am running it on my machine, `make clean` will remove my private SSH key (i.e. ~/.ssh/id_rsa).

dincho: I agree. I forgot to change the naming. Here d577e65

Makefile
```diff
@@ -56,6 +59,17 @@ mnesia_backup:
 		-e backup_suffix=$(BACKUP_SUFFIX) \
 		mnesia_backup.yml
 
+~/.ssh/id_ae_infra_ed25519:
```
lucafavatella: I tested this creates priv and pub key as separate files (id_ae_infra_ed25519 and id_ae_infra_ed25519.pub). Ok.

README.md
```diff
@@ -43,14 +43,19 @@ Vault token could be used by setting `VAULT_AUTH_TOKEN` environment variable (tr
 A Docker image `aeternity/infrastructure` is build and published to DockerHub. To use the image one should configure all the required credentials as documented above and run the container:
 
 ```bash
-docker run -it --env-file env.list -v ~/.ssh/:/root/.ssh/ aeternity/infrastructure
+docker run -it --env-file env.list aeternity/infrastructure
```
lucafavatella: I see VAULT_GITHUB_TOKEN taken from environment via env.list. Ok.

ansible/setup.yml
```diff
@@ -93,6 +57,27 @@
...
         - deploy-ssh-keys
 
+    - name: Download Vault public key
+      get_url:
+        url: "{{ vault_addr }}/v1/ssh-client-signer/public_key"
+        dest: /etc/ssh/trusted-user-ca-keys.pem
+      when: vault_addr is defined
+      tags:
+        - ssh-keys
+        - developer-ssh-keys
```
lucafavatella: This task is only updating the CA cert - not the users. Right?
If I understand correctly, the user SSH-ing is authorized based on GitHub token, that is associated to GitHub account. So it means that user pubkeys on GitHub are not relevant any longer, provided that user provides GitHub token (whether GitHub invalidates tokens when pubkeys changes, that is a separate independent question GitHub-specific). So it means that the list of users is not kept here but in vault (kept and edited directly in vault manually by admin), so there is no need any longer to poll user pubkeys - or any other user info - from GitHub, right?

Makefile
```diff
@@ -56,6 +59,17 @@ mnesia_backup:
...
+	@ssh-keygen -t ed25519 -N "" -f $@
+
+~/.ssh/id_ae_infra_ed25519-%-cert.pub: ~/.ssh/id_ae_infra_ed25519
+	@vault write -field=signed_key ssh/sign/$* public_key=@$<.pub > $@
```
lucafavatella: I assume this line is generating locally CA-signed cert from pubkey after having checked on vault remote server that user to whom the GitHub token is associated is authorized. Right?

Makefile
```diff
@@ -56,6 +59,17 @@ mnesia_backup:
...
+
+ssh-%: ~/.ssh/id_ae_infra_ed25519-%-cert.pub
+	@ssh -i $< -i ~/.ssh/id_ae_infra_ed25519 $*@$(HOST)
+
+ssh: ssh-epoch
```
lucafavatella: I read intended usage as `make ssh HOST=...` after running image attaching to terminal (`-it`). Is user allowed to do make ssh... to a host then exit from that host then run make ssh... for other host? I expect yes. In this case:
* Is there any assumption that user does not present distinct pubkeys/certs in a short succession?
* Is this makefile meant to keep privkey, pubkey and cert between 2 invocations of `make ssh...`? If so, I did not check whether intermediate targets `~/.ssh/id_ae_infra_ed25519-%-cert.pub` and `~/.ssh/id_ae_infra_ed25519` are removed at end of first invocation (while pubkey `*.pub` is for sure kept).

ansible/setup.yml
dincho: Your got it right at 100% !

Makefile
dincho: right, you can read more about this method here: https://code.fb.com/security/scalable-and-secure-access-with-ssh/

However, they use their own tooling instead of vault

Makefile
dincho: > * Is there any assumption that user does not present distinct pubkeys/certs in a short succession?

I'm sorry, I don't get this one, can you rephrase please?

dincho: > * Is this makefile meant to keep privkey, pubkey and cert between 2 invocations of `make ssh...`? If so, I did not check whether intermediate targets `~/.ssh/id_ae_infra_ed25519-%-cert.pub` and `~/.ssh/id_ae_infra_ed25519` are removed at end of first invocation (while pubkey `*.pub` is for sure kept).

The cert is removed, however it's more a side effect. The makefile keeps the keys to slightly reduce the connecting time. The certificate itself have expiration time, the user is free to reuse it until it expires.

lucafavatella: > > * Is there any assumption that user does not present distinct pubkeys/certs in a short succession?
> 
> I'm sorry, I don't get this one, can you rephrase please?

* User runs `make ssh...` for accessing a host.
* User exits from that host (but stays in Docker container).
* User runs `make ssh...` for accessing a host (same as before, or distinct).

I understand from your other answer ("The cert is removed ... The makefile keeps the keys ...") that the cert is removed before the second `make ssh...` so a new cert is generated. My question was about any throttling/restriction that would prevent accessing multiple hosts, e.g. running script in container accessing 20 hosts one after another would attempt to generate 20 certs.

* Case A: If number of generated certs throttled/restricted, multiple `make ssh...` would fail.
* Case B: The accessed host (not the Docker container, but the host to which user SSHs into) grants access to user on the basis on multiple distinct certs. I assume there is no throttling/limit to how many distinct certs the host grants access to.

dincho: I'm not aware of any  throttling/limits so far.

PR #146, opened 2018-11-06
Google cloud platform secrets support

dincho:


import-secrets.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 AWS_CREDS_ROLE="${AWS_CREDS_ROLE:-epoch-fleet-manager}"
+CREDS_ROLE="${CREDS_ROLE:-$AWS_CREDS_ROLE}"
```
dincho: This var duplicating/naming is for backward compatibility

docker-entrypoint.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 
-source import-secrets.sh
+source /infrastructure/import-secrets.sh
```
dincho: Some usages involve changed working directory, thus the absolute path

PR #148, opened 2018-11-07
Add ec2_tag credentials BC workaround

dincho:
This should fix failing setup_monitoring playbook.

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -23,6 +23,12 @@
         resource: "{{ ansible_ec2_instance_id }}"
         state: list
         region: "{{ ansible_ec2_placement_region }}"
+        # Backward compatibility for boto2/botocore because of session token.
+        # Maybe https://github.com/ansible/ansible/pull/43350 will fix that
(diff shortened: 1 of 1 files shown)


ansible/monitoring.yml
```diff
@@ -23,6 +23,12 @@
         resource: "{{ ansible_ec2_instance_id }}"
         state: list
         region: "{{ ansible_ec2_placement_region }}"
+        # Backward compatibility for boto2/botocore because of session token.
+        # Maybe https://github.com/ansible/ansible/pull/43350 will fix that
+        # AWS_SECURITY_TOKEN does not work either ?!
+        aws_access_key: "{{ lookup('env','AWS_ACCESS_KEY_ID') }}"
```
lucafavatella: I see env variables exported in import-secrets.sh
Ok.

lucafavatella: I see ec2_tag properties defined in https://docs.ansible.com/ansible/latest/modules/ec2_tag_module.html
Makes sense.

PR #149, opened 2018-11-08
Fix deprecation warnings in Ansible 2.7. PT-161748227

valkov-ivaylo:


dincho:
@valkov-ivaylo How hard is to add CI check for deprecation warnings ?

valkov-ivaylo:
> @valkov-ivaylo How hard is to add CI check for deprecation warnings ?

I don't know. The warnings probably do not give meaningful return codes. So it must be up to parsing the output. I am not aware if CircleCI has any support for warning detection at all.

PR #153, opened 2018-11-09
Make ops life better

acidsennui:


dincho:
This will easy the mates that work on multiple projects

PR #158, opened 2018-11-14
Bootstrap on reboot

acidsennui:
PR: https://www.pivotaltracker.com/story/show/161835647
refers to: https://github.com/aeternity/epoch-machine-image/pull/10

merged with: https://github.com/aeternity/infrastructure/pull/155

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -44,13 +44,12 @@ PKCS7=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/pkcs7 | 
 export VAULT_ADDR=$vault_addr
 if [ -f "/root/.vault_nonce" ] ; then
     export NONCE=$(cat /root/.vault_nonce)
-
 else
     export NONCE=$(vault write auth/aws/login pkcs7=$PKCS7 role=$vault_role | grep token_meta_nonce | awk '{print $2}')
     echo $NONCE > /root/.vault_nonce
 fi
 
-export VAULT_TOKEN=$(vault write -field=token auth/aws/login pkcs7=$PKCS7 role=$vault_role)
+export VAULT_TOKEN=$(vault write -field=token auth/aws/login pkcs7=$PKCS7 role=$vault_role nonce=$NONCE)
 
 
 
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -30,9 +30,6 @@ case $i in
 esac
 done
 
-
-pwd
-
 cd $(dirname $0)/../ansible/
 
 # Temporary fix/workaround for non-executable vault install
@@ -50,8 +47,6 @@ else
 fi
 
 export VAULT_TOKEN=$(vault write -field=token auth/aws/login pkcs7=$PKCS7 role=$vault_role nonce=$NONCE)
-
-
 
 export env=$env
 export epoch_package=$epoch_package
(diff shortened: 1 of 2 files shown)


scripts/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+
+# Authenticate the instance to CSM
+PKCS7=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/pkcs7 | tr -d '\n')
+
+export VAULT_ADDR=$vault_addr
+if [ -f "/root/.vault_nonce" ] ; then
+    export NONCE=$(cat /root/.vault_nonce)
+else
+    export NONCE=$(vault write auth/aws/login pkcs7=$PKCS7 role=$vault_role | grep token_meta_nonce | awk '{print $2}')
+    echo $NONCE > /root/.vault_nonce
```
dincho: this will use default `umask`, let's make sure the file perms is 0600

scripts/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+fi
+
+export VAULT_TOKEN=$(vault write -field=token auth/aws/login pkcs7=$PKCS7 role=$vault_role nonce=$NONCE)
+
+export env=$env
```
dincho: Can you remind me why this needs to be exported ?

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -1,45 +1,16 @@
...
-    -i /tmp/local_inventory \
-    --become-user epoch -b \
-    -e package=${epoch_package} \
-    -e env=${env} \
-    -e db_version=0 \
-    deploy.yml
+if [ ! -d "/infrastructure" ] ; then
+    git clone -b ${bootstrap_version} --single-branch https://github.com/aeternity/infrastructure.git /infrastructure
+else
+    cd /infrastructure && git pull origin ${bootstrap_version}
```
dincho: This is inconsistent, it enters the `/infrastructure` directory only on reboots ?!

scripts/bootstrap.sh
acidsennui: magic they do not work without this, at least when i was trying to just use env

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
acidsennui: it's not going there in first run because dir do not exists, it performing clone then

scripts/bootstrap.sh
acidsennui: added

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
dincho: Yes, but still it leaves the environment in different state (thus the awkward results in your other PR).

We should either do:
```
git clone ...
cd /infrastructure
```

or 

```
(cd /infrastructure && git pull origin ${bootstrap_version})
```

dincho: Alternatively and cleanest `git -C /infrastructure pull origin ${bootstrap_version}`

acidsennui:
not to my self fix iwth: https://github.com/aeternity/infrastructure/blob/master/import-secrets.sh#L31

acidsennui:
@dincho i think 2 passes now is best way now since first i;m checking is NONCE exist, if yes, i'm using it, if yes, using this in login command, if not logging in saving and reloggign.

other way would be login, check was it success then check is nonce exist and use it for login.

PR #159, opened 2018-11-15
Use custom Docker image entrypoint with CircleCI. PT-161920679 

valkov-ivaylo:


Dockerfile
```diff
@@ -41,5 +41,6 @@ RUN cd /infrastructure/ansible && ansible-galaxy install -r requirements.yml
 ADD . /infrastructure
 WORKDIR /infrastructure
 
+LABEL com.circleci.preserve-entrypoint=true
```
lucafavatella: This looks like a Circle CI feature https://circleci.com/docs/2.0/custom-images/#adding-an-entrypoint
Ok.

PR #162, opened 2018-11-16
Fix ansible AWS checks since 2.7.0 update

dincho:
`ansible_bios_version` is no longer relevant after Ansible 2.7

- [x] test it

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -21,10 +21,22 @@
       # http://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#special-tags
       tags: [always]
       when: hosts_group is defined
+
+    - name: Check if running on AWS instance
+      uri:
+        url: http://169.254.169.254/latest/meta-data
+        timeout: 2
+      register: aws_uri_check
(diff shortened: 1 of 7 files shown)


ansible/deploy.yml
```diff
@@ -21,10 +21,22 @@
       # http://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#special-tags
       tags: [always]
       when: hosts_group is defined
+
+    - name: Check if running on AWS instance
+      uri:
+        url: http://169.254.169.254/latest/meta-data
```
lucafavatella: [AWS metadata service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-retrieval) - ok.

Not sure how stable client and network and service are. It seems appropriate to have a few retries here though I did not find easy option in [doc](https://docs.ansible.com/ansible/latest/modules/uri_module.html).

dincho: Not big deal, we can re-run the playbook or reboot the instance (after #158)

PR #163, opened 2018-11-19
Remove pt 161920679 changes docker entrypoint

valkov-ivaylo:
Remove changes in https://www.pivotaltracker.com/n/projects/2152594/stories/161920679
The approach has issues related to the way CircleCI uses Docker. It is not usable for epoch and epoch-machine-images repositories. Confirmed by @dincho.


valkov-ivaylo:
Will be fixed. It did not have the step before the changes were introduced.

PR #166, opened 2018-11-21
Use Python3 isntead of Python2

dincho:
After long debugging of weird ansible behaviour of copy_module that create binary files with unexpected contents, I found out it's because of the python version. I didn't dig deeper what's the exact reason because python2 should be dead already.

https://www.pivotaltracker.com/story/show/162123459


dincho:
merged into #168

PR #175, opened 2018-11-23
Pin mainnet node version

dincho:


terraform/main.tf
```diff
@@ -51,7 +51,7 @@ module "aws_deploy-main-ap-southeast-1" {
   ami_name      = "epoch-ubuntu-16.04-v1542910070"
 
   epoch = {
-    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-builds/epoch-latest-ubuntu-x86_64.tar.gz"
+    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-1.0.0-rc2-ubuntu-x86_64.tar.gz"
```
lucafavatella: May need tag update though as 1.0.0-rc2 is permanent link it shall work ok.

PR #187, opened 2018-11-28
Allow deployment of just epoch configuration

dincho:


dincho:
> Does this cater for peers?

It should, but however it should not be used on environments with peers autoconfiguration, because .. well you configure the peer keys, so the environment most probably uses static peers (e.g. uat, main .. )



> Is it possible to hook this in smoke test in CI that checks that,

I don't want to overload the tests, if you run such a test for this tag, we should run for all tags combinations ?

PR #190, opened 2018-12-01
Pt 162206445 list inventory

valerifilipov:
Get a grouped list of ansible inventory. Seed nodes are marked with `*`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,10 +145,19 @@ ifndef DEPLOY_ENV
 	$(error DEPLOY_ENV is undefined)
 endif
 
+dump-inventory:
+ifeq ($(wildcard ansible/inventory-list.json),)
+	cd ansible && ansible-inventory --list > inventory-list.json
+endif
+
+list-inventory: dump-inventory
+	cd ansible &&\
+	cat inventory-list.json | ./dump_inventory.py			    
+
 clean:
 	rm ~/.ssh/id_ae_infra*
 
 .PHONY: \
 	images setup-terraform setup-node setup-monitoring setup \
 	manage-node reset-net lint cert-% ssh-% ssh clean \
-	check-seed-peers check-deploy-env
+	check-seed-peers check-deploy-env dump-inventory list-inventory
--- a/Makefile
+++ b/Makefile
@@ -145,19 +145,21 @@ ifndef DEPLOY_ENV
 	$(error DEPLOY_ENV is undefined)
 endif
 
-dump-inventory:
-ifeq ($(wildcard ansible/inventory-list.json),)
+ansible/inventory-list.json:
 	cd ansible && ansible-inventory --list > inventory-list.json
-endif
 
-list-inventory: dump-inventory
+list-inventory: ansible/inventory-list.json
 	cd ansible &&\
-	cat inventory-list.json | ./dump_inventory.py			    
+	cat inventory-list.json | ./dump_inventory.py
 
-clean:
+clean-inventory:
+	rm -f ansible/inventory-list.json
+
+clean: 
 	rm ~/.ssh/id_ae_infra*
+	rm -f ansible/inventory-list.json
 
 .PHONY: \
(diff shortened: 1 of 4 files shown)


ansible/dump_inventory.py
```diff
@@ -32,5 +33,6 @@ def natural_sort_key(s, _nsre=re.compile('([0-9]+)')):
         host_ip = host
         if 'ansible_host' in hostvars[host]:
             host_ip = hostvars[host]['ansible_host']
-        print("- %s (%s) [status](http://%s:3013/v2/status) [top](http://%s:3013/v2/blocks/top)" % (host, host_ip, host_ip, host_ip))
+        bullet = '*' if 'kind' in hostvars[host]['tags'] and hostvars[host]['tags']['kind'] == 'seed' else '-'
```
dincho: How this render in markdown ? The output is and should be kept as valid markdown. Once rendered, users should be able to differentiate between the two type of node.

Alternative implementation would be subgroups: seeds, peers, this implementation is good tho, given the above concerns are not valid.

Makefile
```diff
@@ -145,10 +145,19 @@ ifndef DEPLOY_ENV
 	$(error DEPLOY_ENV is undefined)
 endif
 
+dump-inventory:
```
dincho: Isn't this the equivalent (mostly) to ?:
```
ansible/inventory-list.json:
 	cd ansible && ansible-inventory --list > inventory-list.json
```

Let's try to use the makefile features where possible, I know we're abusing it, but still

README.md
```diff
@@ -66,6 +66,16 @@ Check that your AWS credentials are setup and dynamic inventory is working as ex
...
 
+### List inventory
+
+Get a grouped list of ansible inventory. Seed nodes are marked with `*`
+
+```
+make list-inventory
+```
+
+Inventory data is stored in local file `ansible/inventory-list.json`. To refresh it later you can `make dump-inventory`
```
dincho: I don't think the second sentence is valid, isn't the current implementation only dump it if the file does not exists yet?

Makefile
valerifilipov: Yes, it's the better way, i chose the other way because of shorter name, will change it

ansible/dump_inventory.py
valerifilipov: Sorry, never thought output actually should be in some valid format, will change it

ansible/dump_inventory.py
```diff
@@ -24,13 +24,27 @@ def natural_sort_key(s, _nsre=re.compile('([0-9]+)')):
...
     for host in data[group_name]['hosts']:
-        host_ip = host
-        if 'ansible_host' in hostvars[host]:
-            host_ip = hostvars[host]['ansible_host']
-        print("- %s (%s) [status](http://%s:3013/v2/status) [top](http://%s:3013/v2/blocks/top)" % (host, host_ip, host_ip, host_ip))
+        if 'kind' in hostvars[host]['tags'] and hostvars[host]['tags']['kind'] == 'seed':
+            seeds.append(host)
+        else:
+            peers.append(host)
+    print('seeds:')
```
dincho: let's add some bold formatting to this sections:
```suggestion
    print('**seeds:**')
```

ansible/dump_inventory.py
```diff
@@ -24,13 +24,27 @@ def natural_sort_key(s, _nsre=re.compile('([0-9]+)')):
...
+    for seed in seeds:
+        print_host(seed)
+    print('peers:')
```
dincho: same as above
```suggestion
    print('**peers:**')
```

ansible/dump_inventory.py
dincho: let's not print this section header if there are no seed peers in that environment

dincho:
> Could of cosmetic changes, otherwise it's working as expected.

s/could/couple

PR #191, opened 2018-12-03
Bump max_gossip to 1024 on main nodes

dincho:


ansible/vars/epoch/main.yml
```diff
@@ -87,7 +87,7 @@ epoch_config:
 
   sync:
     port: 3015
-    max_gossip: 100
+    max_gossip: 1024
```
lucafavatella: I understand to be reduced in the future.

PR #192, opened 2018-12-03
Run tests in containerized python3 environment

dincho:
https://www.pivotaltracker.com/story/show/162164555

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,6 @@ jobs:
             source import-secrets.sh
             docker login -u $DOCKER_USER -p $DOCKER_PASS
             docker push ${DOCKERHUB_REPO:?}:${CIRCLE_BRANCH:?}
-      - *fail_notification
 
   setup_terraform:
     <<: *container_config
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -85,53 +85,57 @@ references:
 
 jobs:
   build_check:
-    <<: *machine_config
+    docker:
+      - image: aeternity/infrastructure:master
+      - image: rastasheep/ubuntu-sshd:16.04
```
valkov-ivaylo: Mainly for security reasons I have reservations related to third party containers from unverified sources. Can this be trusted? What is the issue that requires the second image? Why the steps cannot run in a single container? I am aware that Docker is not welcoming sane solutions. If it is just the sshd service, is there an easy, way to add it in the infrastructure image without that causing major issues to the operation of the container?

dincho: Our infrastructure container uses alpine, while we deploy on Ubuntu, thus we have to test on ubuntu, thus the ubuntu-sshd container.

While I understand your concerns, using a 3rd party container is the same as using any 3rd party library, the same pros/cons.

Furthermore we don't share any secrets with the container during the test, so I think we're relatively *safe*.

valkov-ivaylo: I do not agree completely with the library analogy, but I'll accept it.

PR #194, opened 2018-12-04
Bump mainnet nodes version to 1.0.1

dincho:


terraform/main.tf
```diff
@@ -57,7 +57,7 @@ module "aws_deploy-main-ap-southeast-1" {
   ami_name      = "epoch-ubuntu-16.04-v1542910070"
 
   epoch = {
-    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-1.0.0-ubuntu-x86_64.tar.gz"
+    package = "https://s3.eu-central-1.amazonaws.com/aeternity-epoch-releases/epoch-1.0.1-ubuntu-x86_64.tar.gz"
```
lucafavatella: This is used by `terraform init` and `terraform apply` triggered on CI on master merge, that pushes config to amazon so on spot instance bootstrap the expected package variable is passed. Ok.

PR #198, opened 2018-12-10
Change beneficiary for the UAT nodes

noandrea:


ansible/vars/epoch/uat.yml
```diff
@@ -20,6 +20,7 @@ epoch_config:
       port: 3013
     internal:
       port: 3113
+    debug: true
```
velzevur: I see this is actually [this](https://github.com/aeternity/epoch/blob/master/apps/aeutils/priv/epoch_config_schema.json#L281) and not the tag `debug`? I am digging deeper

velzevur: Regarding this, I expect a behavior change introduced in aeternity/epoch#1901

noandrea: it is a bit confusing to understand the difference between [this](https://github.com/aeternity/epoch/blob/e8051ab20352a25727d15d3503c8640792258dca/apps/aeutils/priv/epoch_config_schema.json#L272) and [this](https://github.com/aeternity/epoch/blob/e8051ab20352a25727d15d3503c8640792258dca/apps/aeutils/priv/epoch_config_schema.json#L281) from the description. 

Maybe @fabiankrol can suggest what is the right way to enable the debug/internal endpoints?

ansible/vars/epoch/uat.yml
```diff
@@ -30,7 +31,7 @@ epoch_config:
     db_path: "./db{{ db_version|mandatory }}"
 
   mining:
-    beneficiary: "ak_tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
+    beneficiary: "ak_2iBPH7HUz3cSDVEUWiHg76MZJ6tZooVNBmmxcgVK6VV8KAE688"
```
lucafavatella: 👍

ansible/vars/epoch/uat.yml
lucafavatella: https://github.com/aeternity/epoch/pull/1901

noandrea: due to ongoing changes and other impacts, let's take out the `debug` change so the pr can be merged

PR #199, opened 2018-12-10
Improve SSH authentication docs

dincho:
https://www.pivotaltracker.com/story/show/162341740


README.md
```diff
@@ -12,53 +12,108 @@ Below documentation is meant for manual testing and additional details. It's alr
...
-AWS_PROFILE=aeternity ansible-inventory --list
+export VAULT_ADDR=https://the.vault.address/
+export VAULT_GITHUB_TOKEN=your_personal_github_token
+```
+
+Run the container:
+
+```
+docker pull aeternity/infrastructure
+docker run -it --env-file env.list aeternity/infrastructure
```
valkov-ivaylo: It is a bit unclear from where the `docker run` command should be ran. The assumption is that it should be ran from within a copy of the infrastructure repository. However, reading the documentation and following the steps, one might get carried away, and just follow the steps without cloning. It should be clear that `docker run` must be executed from a copy of the infrastructure repository, so the env.list file is present. 

My suggestion is to add the following two steps before the `docker run` command, so it is obvious. Especially to new team members:

`git clone git@github.com:aeternity/infrastructure.git`
`cd infrastructure`

Either that, or steps to create the env.list file manually.

dincho: Good point, I forgot to work on that issue. I updated the commands in ce963ae

PR #200, opened 2018-12-13
Enable mining on all environments but main

dincho:
https://www.pivotaltracker.com/story/show/162634314

### DIFF
--- a/ansible/vars/epoch/main.yml
+++ b/ansible/vars/epoch/main.yml
@@ -105,7 +105,6 @@ epoch_config:
 
   mining:
     autostart: false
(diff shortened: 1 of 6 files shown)


ansible/vars/epoch/main.yml
```diff
@@ -105,7 +105,6 @@ epoch_config:
 
   mining:
     autostart: false
-    beneficiary: "ak_tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5"
```
zp-sd: `autostart: false` above is redundant

dincho: I know but I don't want to rely on defaults ;)

PR #203, opened 2018-12-18
PT-162445499 Change cuckoo config to allow multiple simultaneous miners

zp-sd:
Depends on https://github.com/aeternity/epoch/pull/1942.


zp-sd:
Given https://github.com/aeternity/epoch/pull/1942 makes new config backwards compatible now, this can be merged even after https://github.com/aeternity/epoch/pull/1942 is merged & released.

zp-sd:
Integration env checked - chain still growing.

PR #209, opened 2019-01-17
Increase disk size

dincho:
- on-demand main nodes configuration has been applied manually region-by-region.
- all main spot instances configuration has been applied at once
- merging this PR will apply some configuration changes (because of 
002637c809d6eb800bb576b6c150ce2f040d16d)2 to the development environments (uat, dev*, integration), however downtime is not expected neither functionality changes (no disk space increase)

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -11,25 +11,25 @@ variable "vault_addr" {
 }
 
 provider "aws" {
-  version = "1.24"
+  version = "1.55"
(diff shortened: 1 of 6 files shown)


dincho:
@lucafavatella Mainnet nodes has been already recreated and resync yesterday. This PR is technically to sync (inverse sync) the terraform state (S3) with the code.

PR #210, opened 2019-01-17
Rename all variables bc pt 163189722

valerifilipov:


dincho:
please rebase on master

terraform/modules/cloud/aws/deploy/fleet/sg.tf
```diff
@@ -64,7 +64,7 @@ resource "aws_security_group_rule" "allow_ssh" {
   security_group_id = "${aws_security_group.ae-nodes-management.id}"
 }
 
-resource "aws_security_group_rule" "allow_epoch_icmp" {
+resource "aws_security_group_rule" "allow_aeternity_icmp" {
```
dincho: I would use more neutral name here, like `allow_icmp`

dincho:
Could you also please change the `tag_role_epoch` (and corresponding TF config) to `tag_role_aenode`

dincho:
> I see S3 bucket and package name not renamed. Ok.

it's tracked in https://www.pivotaltracker.com/story/show/163189736

PR #214, opened 2019-01-22
Change aeternity nodes role name

dincho:
In scope of https://www.pivotaltracker.com/story/show/163323269

While 03e8897 will not cause instance re-spawn, 1cac4e4 will do because it changes the bootstrap user data, thus the main nodes are not changed for now.


dincho:
Build fails because the CI is configured after https://github.com/aeternity/infrastructure/pull/213 and it's not merged yet.

PR #218, opened 2019-01-24
Change nodes username. PT-163298922

valkov-ivaylo:
Updated:
 * Ansible playbooks
 * documentation
 * Makefile
 * bootstrap.sh
 * Ansible fact file

Should **NOT** be merged before 1.3.0 is deployed.


valkov-ivaylo:
> After daily meeting additional changes were requested. An Ansible playbook will be created to migrate usernames on running nodes.

Added in 68d59de. Tested on local machine. It is not called by anything else. Should be run manually.

ansible/rename-user-on-old-nodes.yml
```diff
@@ -0,0 +1,43 @@
+---
```
dincho: isn't not very ansible is it? I'd expect idempotent playbook that runs commands only when necessary. What happens if you rerun this playbook ?

valkov-ivaylo: Idempotency probably could be fixed. The shell commands are necessary. Ansible modules do not support user and group renaming (keep id, move home directory, set permissions etc.). At least it is not obvious how from the documentation.

Deleting users and groups, creating new ones, moving/copying files and changing permissions to new ones with modules and tasks seems like an overkill. Deleting users especially is never a good idea if one simply wants to change the username, but not the ID. We do not want to change the gid and uid, because of the list of tasks required and the complexity to do it with an abstract layer such as Ansible. Using the "rename" capabilities of the OS (usermod, groupmod) is much simpler.

valkov-ivaylo: Should be fixed in 2ab740e . Keeping the UID turned out to be error-prone. The permissions of the moved home directory are changed to the new user.

dincho:
Don't forget to `squash & merge` !

PR #227, opened 2019-01-29
Change uat instance type and spot instance price

valerifilipov:
re https://www.pivotaltracker.com/story/show/163059780


valerifilipov:
This commit should recreate launch configuration for uat spot instances and launch them in Stockholm with new type m5.large and spot price. Also launch one static uat node. Previous launch on #224 merge failed due to unsupported instance type m4.large in this region

PR #231, opened 2019-01-30
Change integrations tests region to Paris

valerifilipov:
Re https://www.pivotaltracker.com/story/show/163511600


dincho:
Do we have Vault and Packer configurations in place already ?

valerifilipov:
> Do we have Vault and Packer configurations in place already ?

No, manually copied lastest ami

PR #239, opened 2019-02-05
change ec2 node size

michalbagrowski:
Because of node size do not give us big difference with synch time, decreasing to smallest / cheapest instance with 2gb of mem.

screenshot 2019-02-05 12 55 22

screen show sync time from 20400 -> 33200 in seconds(3rd row) / minutes (4th row)

PT: https://www.pivotaltracker.com/story/show/163018427

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -83,7 +83,7 @@ module "aws_deploy-main-eu-west-2" {
   spot_nodes   = 0
 
   spot_price       = "0.15"
-  instance_type    = "c5.xlarge"
+  instance_type    = "t3.small"
   ami_name         = "epoch-ubuntu-16.04-v1542910070"
   root_volume_size = 16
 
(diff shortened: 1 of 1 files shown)


terraform/main.tf
```diff
@@ -184,7 +184,7 @@ module "aws_deploy-ap-southeast-1" {
   spot_nodes   = 14
 
   spot_price    = "0.125"
-  instance_type = "m4.large"
+  instance_type = "t3.small"
```
dincho: We can't mine on this instance, we need 4GB+ memory and more stable CPU ?

dincho:
Please rebase and open PR against `destructive_updates`

dincho:
More context: https://www.pivotaltracker.com/story/show/163730878

terraform/main.tf
michalbagrowski: ok did not know that we have diffrent node types :D so do we need to run some test or just change it to something cheaper with 4gb of mem?

dincho:
> @dincho hm... i think we should split node size for static / non static nodes, what do you think about this?

You mean for main environment? There is no spot nodes there anymore

michalbagrowski:
@dincho no for any env, no point of having static (not mining, as far as i understand) nodes with big instances

PR #240, opened 2019-02-05
fix integration

michalbagrowski:
fix integration env, missing aeternity user for playbooks


michalbagrowski:
@lucafavatella actually closing this, didn't know that there is a playbook to fix this, @valerifilipov  is trying to run it as we speak.

PR #241, opened 2019-02-06
Improve error handling on nonce 163696101

michalbagrowski:
PT: https://www.pivotaltracker.com/story/show/163696101

Tested for empty nonce:

```
master@ip-10-0-1-135:~$ cat /tmp/user-data.log
Cloning into '/infrastructure'...
WARNING! The following warnings were returned from Vault:

  * TTL of "768h0m0s" exceeded the effective max_ttl of "30m0s"; TTL value is
  capped accordingly

NONCE is empty
master@ip-10-0-1-135:~$
```
do not have idea how to test it on saving issue, but this `if` should work

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -44,7 +44,17 @@ if [ -f "/root/.vault_nonce" ] ; then
     export NONCE=$(cat /root/.vault_nonce)
 else
     export NONCE=$(vault write auth/aws/login pkcs7=$PKCS7 role=$vault_role | grep token_meta_nonce | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


michalbagrowski:
@dincho i have tested empty nonce, just was not able to test second one `if [ $? -nq 0 ]; then` 

but if this is to much should i just leave it there or remove?

dincho:
> but if this is to much should i just leave it there or remove?

I really don't see use case and given it cannot even be tested, let's remove it.

michalbagrowski:
@dincho removed.

PR #242, opened 2019-02-06
Force start node in playbook 163698014

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/163698014

Do not see reason to have any conditions there.

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -119,10 +119,6 @@
             # allow group read for integration purposes  (e.g. logs)
             mode: 0750
 
-      - name: Start aeternity node
-        command: "{{ project_root }}/bin/aeternity start"
-        tags: [daemon]
-
       - meta: flush_handlers
 
       - name: Create erlang cookie file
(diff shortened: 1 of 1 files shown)


ansible/deploy.yml
```diff
@@ -204,7 +204,8 @@
       - name: Trigger start ae node
         command: "/bin/true"
         notify: "restart aeternity daemon"
```
dincho: Why restart? We need start, otherwise running the playbook will always restart the node, this is not what we want

ansible/deploy.yml
```diff
@@ -204,7 +204,8 @@
       - name: Trigger start ae node
```
dincho: I find `Ensure node is running` a better name, while not 100% correct

michalbagrowski:
Will fix, leftover form trying to hack it before.

czw., 7 lut 2019, 11:30 użytkownik Dincho Todorov <notifications@github.com>
napisał:

PR #243, opened 2019-02-06
Rename epoch leftovers 163730878

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/163730878

all done tested with new env setup.

change vault_addr will be done on manual update step by step, will have to remember after to update circleci env variable.


dincho:
> change vault_addr will be done on manual update step by step, will have to remember after to update circleci env variable.

good point

PR #244, opened 2019-02-06
Fix OS user renaming issues. PT-163298922

valkov-ivaylo:
Fixed issues with: 
 * undying epmd processes
 * foreign users in epoch group
 * added checks for existence of old group
 * switched to usermod and groupmod

Tested against dev1 and integration.

### DIFF
--- a/ansible/rename-user-on-old-nodes.yml
+++ b/ansible/rename-user-on-old-nodes.yml
@@ -76,6 +76,7 @@
         recurse: yes
         file_type: file
       register: epmd_path
+      become: yes
 
     - name: Stop epmd leftover process
       command: "{{ epmd_path.files[0].path }} -kill"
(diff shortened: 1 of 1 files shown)


ansible/rename-user-on-old-nodes.yml
```diff
@@ -23,51 +22,55 @@
...
 
-    - name: Check new home directory
-      stat:
-        path: "{{ project_new_home }}"
-      register: new_home_present
+    - name: Find epmd binary
+      find:
+        paths:
+          - "{{project_old_home}}"
+          - "{{ project_new_home }}"
```
michalbagrowski: one var have spaces one do not have ;] just looks not nice ;]

valkov-ivaylo: Fixed in 4a42c1f.

michalbagrowski: @valkov-ivaylo yea... can you make this consistent for whole playbook? :D it looks much beter with spaces, and we use spaces everywhere {{ something }}

PR #250, opened 2019-02-12
fix unstable peer keys

valerifilipov:
wrong seed pubkey for unstable env
redeployed with new config


dincho:
Also please add a Makefile check

PR #252, opened 2019-02-13
Manual Vault token refresh (Increase token timeouts). PT-163087449

valkov-ivaylo:
Add script to manually refresh Vault tokens (and leases). 
Updated README file regarding refreshing Vault secrets.

### DIFF
--- a/refresh-secrets.sh
+++ b/refresh-secrets.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+vault token revoke -self
+unset VAULT_TOKEN
+source /infrastructure/import-secrets.sh
--- a/refresh-secrets.sh
+++ b/refresh-secrets.sh
@@ -2,4 +2,4 @@
 
 vault token revoke -self
 unset VAULT_TOKEN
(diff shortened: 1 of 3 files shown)


refresh-secrets.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+vault token revoke -self
+unset VAULT_TOKEN
+source /infrastructure/import-secrets.sh
```
dincho: This should be a relative path. I often use the container with `/src` mounted.

valkov-ivaylo: From my understanding `/infrastructure` should always be present. Relative to what? Current working directory could be any. The source will not find the file then. What am I missing?

dincho: During the development I use `/src` mount for example, if this is kept as is, I need to rebuild the container each time I change the `import-secrets.sh` to test the `refresh-secrets.sh`

valkov-ivaylo: Fixed.

PR #253, opened 2019-02-14
decrease donwtime with node rotation

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/163826822

So for static nodes:

1. tf will create boot new node
2. tf will swap IP address to new node. No health check it just wait for instance boot, and swap ip.

* if terrafrom will detect that changes do not require instance termination, it will stop / start node like just for changing instance size.
For spot nodes:

1. tf will create new ASG run new nodes
2. tf will terminate old nodes


michalbagrowski:
make sure that this will not remove IP from existing ENV.

michalbagrowski:
OK so this change will rotate `non static nodes` nad will keep static nodes witch keeping IP when i.e. ami change will be required. tested on new env. cc @dincho

PR #255, opened 2019-02-15
Ansible async provision playbook pt 162699043

valerifilipov:
https://www.pivotaltracker.com/story/show/162699043
A playbook to run remotely bootstrap.sh on nodes asynchronously to prevent ansible ssh session timeout.
Could be run from infra container like:
```bash
ansible-playbook --limit="tag_env_${DEPLOY_ENV}:&tag_role_aenode" \
    -e ansible_python_interpreter=/usr/bin/python3 \
    -e env=${DEPLOY_ENV} \
    -e vault_addr=${VAULT_ADDR} \
    -e package=${PACKAGE} \
    -e bootstrap_version=master
    bootstrap.yml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,6 +170,15 @@ list-inventory: ansible/inventory-list.json
 	cd ansible &&\
 	cat inventory-list.json | ./dump_inventory.py
 
+provision: check-deploy-env
+	cd ansible && ansible-playbook --limit="tag_env_$(DEPLOY_ENV):&tag_role_aenode" \
+	-e ansible_python_interpreter=/usr/bin/python3 \
+	-e env=$(DEPLOY_ENV) \
+	-e vault_addr=$(VAULT_ADDR) \
+	-e package=$(PACKAGE) \
+	-e bootstrap_version=$(BOOTSTRAP_VERSION) \
+	async_provision.yml
+
 clean:
 	rm ~/.ssh/id_ae_infra*
 	rm -f ansible/inventory-list.json
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,15 @@ mnesia_backup:
 		-e backup_suffix=$(BACKUP_SUFFIX) \
 		mnesia_backup.yml
 
+provision: check-deploy-env
+	cd ansible && ansible-playbook --limit="tag_env_$(DEPLOY_ENV):&tag_role_aenode" \
+	-e ansible_python_interpreter=/usr/bin/python3 \
+	-e env=$(DEPLOY_ENV) \
+	-e vault_addr=$(VAULT_ADDR) \
+	-e package=$(PACKAGE) \
+	-e bootstrap_version=$(BOOTSTRAP_VERSION) \
+	async_provision.yml
+
 ~/.ssh/id_ae_infra_ed25519:
 	@ssh-keygen -t ed25519 -N "" -f $@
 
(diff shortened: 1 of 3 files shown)


ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
...
+        depth: 1
+        force: yes
+
+    - name: Run bootstrap.sh
+      shell: |
+        /infrastructure/scripts/bootstrap.sh \
+         --vault_addr={{ vault_addr }} \
+         --vault_role={{ vault_role }} \
+         --env={{ env }} \
+         --epoch_package={{ package }} \
```
dincho: rebase and sync the name with master

ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
...
+         >> /tmp/user-data.log 2>&1
```
dincho: I don't think it's good to append to that file. I'm not sure if there a better way to handle the output at all.

ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
...
+      async: 900
```
dincho: is this 15 minutes? Isn't it too much, what's the runtime on new instance ?

ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
+---
```
dincho: Let's use the term "provision". I know it may be confusing because of boostrap.sh, but that was used to bootstrap only at the beginning. Now we shall change the naming a bit (other PR for the .sh and user_data).

Also let's add the "async" in the name so `async_provision.yml` sounds good to me?

*also the playbook name"

ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Bootstrap node servers
+  hosts: all
+  remote_user: master
+  become: yes
+
+  vars:
+    bootstrap_version: master
```
dincho: This should be `stable`

ansible/bootstrap.yml
```diff
@@ -0,0 +1,31 @@
...
+    vault_role: ae-node
+    package: "https://s3.eu-central-1.amazonaws.com/aeternity-node-builds/aeternity-latest-ubuntu-x86_64.tar.gz"
```
dincho: I prefer to have the `latest` stable package by default instead of latest unknown build.

ansible/bootstrap.yml
valerifilipov: Then maybe append to /tmp/provision.log ? I think it's good to have the output somewhere, because when running the playbook you see nothing, just the status. Another thing i will try is to register output and display it during the play on fail

ansible/bootstrap.yml
valerifilipov: I'm not sure about 'async' in the name, because actually we don't have that with just one task

ansible/bootstrap.yml
valerifilipov: AFAIR it's less than 5 mins, i thought it's not good to fail because of slow download. I'll lower it

ansible/bootstrap.yml
valerifilipov: fixed https://github.com/aeternity/infrastructure/pull/255/commits/4bee91dd652f35af5c5c3318671d9bbfcf9da9a8

Makefile
```diff
@@ -170,6 +170,15 @@ list-inventory: ansible/inventory-list.json
 	cd ansible &&\
 	cat inventory-list.json | ./dump_inventory.py
 
+provision: check-deploy-env
```
dincho: Could you please move it up in the file where all other playbooks are set?

What happens if the BOOTSTRAP_VERSION is not set?

valerifilipov: defaults to what's in playbook, e.g. `stable` https://github.com/aeternity/infrastructure/blob/b093dd58017b5598daf94e32c96a6edb00cb2cee/ansible/async_provision.yml#L8

PR #258, opened 2019-02-21
Add goss infrastructure docker pt 163087220

valerifilipov:
https://www.pivotaltracker.com/story/show/163087220
* add goss to infrastructure container
* add test check for seed nodes pubkeys


valerifilipov:
> * source the peers from ansible/vault
> * able to run the checks for a specified env only (also for all)
i think this one should check for what's in node configs, like it is right now, and add other targets, which source from ansible end per env, or whatever
> * don’t track the list? Or maybe it’s better to track it?
i think this specific list shouldn't be tracked as it is autogenerated, will see for others and change paths accordingly

dincho:
No, we must not add reverse dependency on the Aeternity node repo.
The dependency should be kept only aeternity-> infrastructure

dincho:
Keeping a tracked list has the huge advantage to catch all integration errors that can popup

valerifilipov:
> No, we must not add reverse dependency on the Aeternity node repo.
> The dependency should be kept only aeternity-> infrastructure

ok then get seeds list from here https://github.com/aeternity/infrastructure/blob/master/ansible/vars/aeternity/main.yml#L46 ?

dincho:
> > No, we must not add reverse dependency on the Aeternity node repo.
> > The dependency should be kept only aeternity-> infrastructure
> 
> ok then get seeds list from here https://github.com/aeternity/infrastructure/blob/master/ansible/vars/aeternity/main.yml#L46 ?

A p2p network != environment. We could have many environments that joins the mainnet. Furthermore we'd probably remove the peer list from this configuration file as it's already hardcoded in the aeternity node. Furthermore this list will include peers that's not under our control in future (community stable peers).

This test should cover only our nodes. I'd propose to have a script that generate a list of our seed nodes and we track the list to catch integration issues, but still to be easy to update the list with a single command.

test/goss/scripts/dump-seed-peers-keys.py
```diff
@@ -0,0 +1,40 @@
...
+    if vproc.returncode > 0:
+        print(err)
+#        sys.exit(1)
+    return out.decode("utf-8")
+
+env_seed_list = []
+for host in inventory[env]['hosts']:
+    if host in inventory['tag_kind_seed']['hosts']:
+        env_seed_list.append({'ip_addr': host, 'pubkey': get_peer_pubkey_from_vault(host)})
+json.dump({'hosts': env_seed_list}, sys.stdout)
```
dincho: pretty print or nicer formatting ?

valerifilipov: may be you commented on a previous commit, i changed it to yaml for easier reading

test/goss/remote/goss.yaml
```diff
@@ -0,0 +1,3 @@
+gossfile:
```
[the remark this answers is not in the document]
valerifilipov: It's the default checks file which `goss` runs if not otherwise specified with `-g`. It can contain multiple gossfiles. I'll remove it as it's not used

test/goss/remote/scripts/dump-seed-peers-keys.py
```diff
@@ -0,0 +1,40 @@
...
+    vproc = subprocess.Popen(vault_cmd, stdout=subprocess.PIPE)
+    out, err = vproc.communicate()
+    if vproc.returncode > 0:
+        print(err, file=sys.stderr)
+#        sys.exit(1)
+    return out.decode("utf-8")
+
+env_seed_list = []
+for host in inventory[env]['hosts']:
+    if host in inventory['tag_kind_seed']['hosts']:
```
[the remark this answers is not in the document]
valerifilipov: sorted

test/goss/remote/scripts/dump-seed-peers-keys.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
valerifilipov: moved

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM alpine:3.8
 # Some of the playbooks run on multiple hosts: local + remote.
 # If it's set in the inventory it will not work when a specific inventory is used.
 # OpenSSL required for a packer workaround: https://github.com/hashicorp/packer/issues/2526
-RUN apk add --no-cache bash curl unzip make python3 py-cryptography openssh-client openssl sshpass \
+RUN apk add --no-cache bash curl unzip make python3 py-cryptography openssh-client openssl sshpass gawk\
```
dincho: I don't see `gawk` usage ?

.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ references:
   check_seed_peers: &check_seed_peers
     run:
       name: Check seed peers
-      command: make check-seed-peers
+      command: make check-seed-peers-all
```
dincho: I guess it's also OK to remove the old command ?

dincho:
Also needs rebase

PR #264, opened 2019-02-25
Speed up initial DB synch

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/163475803


ansible/templates/aeternity.backup.yaml
```diff
@@ -2,8 +2,8 @@
 chain:
   persist: true
   db_path: "./db{{ db_version|mandatory }}"
-{% if node_config['fork_management'] %}
-{% if node_config['fork_management']['network_id'] %}
+{% if node_config['fork_management'] is defined %}
+{% if node_config['fork_management']['network_id'] is defined %}
 fork_management:
     network_id: {{ node_config.fork_management.network_id }}
```
dincho: it should be safe to use `ae_mainnet` as default instead of the ifs

https://github.com/aeternity/aeternity/blob/4fecd35b2b3bf2b311a306d13db22fb20864c496/apps/aecore/src/aec_governance.erl#L229
https://github.com/aeternity/aeternity/blob/4fecd35b2b3bf2b311a306d13db22fb20864c496/apps/aecore/src/aec_governance.erl#L38

ansible/mnesia_backup.yml
```diff
@@ -16,33 +16,66 @@
...
+        bucket: "{{ backup_bucket }}"
+        object: "{{ backup_archive_filename }}"
+        src: /tmp/{{ backup_archive_filename }}
+        mode: put
 
-    - name: Download the Mnesia backup archive
-      fetch:
-        src: /tmp/{{ backup_filename }}.gz
-        dest: "{{ download_dir }}"
+    - name: Upload backup to S3 bucket latest
```
dincho: Isn't it more effective to do a s3 cp of the above object ? Both in term of speed and cost ?

ansible/mnesia_restore.yml
```diff
@@ -0,0 +1,66 @@
+---
+# Playbook used to download (archived) Mnesia backup from last host in the playbook hosts
```
dincho: Please update the docs to be related to this playbook.

ansible/mnesia_restore.yml
```diff
@@ -0,0 +1,66 @@
...
+      unarchive:
+        remote_src: yes
+        src: /tmp/{{ backup_archive_filename_latest }}
+        dest: /tmp/
+
+    - import_tasks: backup/disable_mining_and_peers.yml
+
+    - meta: flush_handlers
+
+    - name: Wait aeternity node API to boot
```
dincho: why you don't include `tasks/health_check.yml` ?

local_run.sh
```diff
@@ -0,0 +1,2 @@
+docker build -t aeternity/local .
+docker run -it --env-file=env.list -v ${PWD}:/src aeternity/local
```
dincho: `-w /src` is also nice option

ansible/backup/disable_mining_and_peers.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Load a variable files for environment: {{ env }}"
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "../vars/aeternity/{{ env }}.yml"
+    - "../vars/aeternity/default.yml"
+  tags: [config, node_config, peer_keys, health-check]
```
dincho: I see unrelated tags

ansible/deploy.yml
```diff
@@ -287,36 +287,4 @@
...
-        - "stop aeternity daemon"
-        - "restart aeternity daemon"
-
-    - name: Start aeternity node
-      command: "{{ project_root }}/bin/aeternity start"
-      tags: [daemon]
-      listen:
-        - "start aeternity daemon"
-        - "restart aeternity daemon"
+  - import_tasks: handlers.yml
```
dincho: please move the file to `tasks/` dir where we keep the partials

ansible/mnesia_restore.yml
michalbagrowski: because this: https://github.com/aeternity/infrastructure/blob/master/ansible/deploy.yml#L227 was using this as source.

ansible/mnesia_backup.yml
michalbagrowski: hm... not sure will aws cp will be smart enought to copy it without downloading, worth to try.

ansible/mnesia_backup.yml
```diff
@@ -16,33 +16,60 @@
...
+      tags: [health-check]
 
-  tasks:
     - name: Backup Mnesia database
-      command: "{{ aeternity_bin }} eval 'ok = mnesia:backup(\"/tmp/{{ backup_filename }}\").'"
+      command: "{{ aeternity_bin }} eval 'ok = mnesia:backup(\"/tmp/backup\").'"
       args:
-        creates: "/tmp/{{ backup_filename }}"
+        creates: "/tmp/backup"
+      notify: "stop aeternity daemon"
```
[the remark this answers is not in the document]
michalbagrowski: hm.. copy & paste error

ansible/templates/aeternity.backup.yaml
michalbagrowski: should or it's safe? and i can use network_id as default for fork_management.network_id but still will fail on fork_management do not exist.

ansible/mnesia_backup.yml
michalbagrowski:
```get
put
delete
create
geturl
getstr
delobj
list```


module do not have CP option

File: ansible/mnesia_restore.yml (hunk 4 above)
Comment (michalbagrowski) [reply ansible/mnesia_restore.yml]: and because VENV configuration 
```TASK [Wait aeternity node API to boot] *****************************************
Tuesday 12 March 2019  11:03:05 +0000 (0:00:00.866)       0:00:16.830 ********* 
fatal: [18.130.244.71]: FAILED! => changed=false 
  module_stderr: |-
    /bin/sh: /var/venv/bin/python: not found
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 127
	to retry, use: --limit @/infrastructure/ansible/mnesia_snapshot.retry```

File: .circleci/config.yml [hunk 9]
```diff
@@ -13,7 +13,7 @@ references:
       docker_layer_caching: true
     working_directory: /home/circleci/infrastructure
     environment:
-      DOCKERHUB_REPO: aeternity/infrastructure
+      DOCKERHUB_REPO: aeternity/infrastructure:master
```
dincho: This looks wrong, not sure what's the reason for.

.circleci/config.yml
```diff
@@ -99,6 +99,7 @@ jobs:
       ANSIBLE_HOST_KEY_CHECKING: False
     steps:
       - checkout
+      - *setup_secrets
```
dincho: I don't see any changes in the steps below, why do you need that? Secrets are only needed and used explicitly here: https://github.com/aeternity/infrastructure/pull/264/files#diff-1d37e48f9ceff6d8030570cd36286a61R145

ansible/tasks/config.yml
```diff
@@ -0,0 +1,9 @@
+project_root: "{{ ansible_env.HOME }}/node"
+aeternity_bin: "{{ ansible_env.HOME }}/node/bin/aeternity"
+download_dir: /tmp/mnesia_backups
+backup_bucket: aeternity-database-backups
+backup_suffix: backup
+health_check: yes
+format: gz
```
valerifilipov: May be this should be **zip** as it is changed in snapshot playbook ?

.circleci/config.yml
michalbagrowski: it was required when there was playbook idempodency test, since we do not need them removed.

ansible/tasks/config.yml
michalbagrowski: snapshoot should use {{ format }} fixed.

Makefile
```diff
@@ -66,13 +66,25 @@ reset-net: check-deploy-env
...
-		-e ansible_python_interpreter=/usr/bin/python3 \
+		-e ansible_python_interpreter=/var/venv/bin/python \
 		-e download_dir=$(BACKUP_DIR) \
 		-e backup_suffix=$(BACKUP_SUFFIX) \
-		mnesia_backup.yml
+		-e db_version=$(BACKUP_DB_VERSION) \
+		-e env=$(BACKUP_ENV) \
+		mnesia_snapshot.yml
+
+mnesia_snapshot_restore:
```
dincho: if this is not working, let's remove it ?

.circleci/config.yml
```diff
@@ -90,7 +90,7 @@ commands:
 jobs:
   build_check:
     docker:
-      - image: aeternity/infrastructure:master
+      - image: aeternity/infrastructure:stable
```
dincho: Not sure if I asked already, but what's the purpose of this ? You can't integrate BC breaking features if this is set to stable, because you need to run at least the latest master.

Makefile
michalbagrowski: Not sure what you arę talking here that is not working

michalbagrowski: Not sure what you arę talking here that is not working

.circleci/config.yml
michalbagrowski: Consider this as typo ;) will fix

michalbagrowski: Consider this as typo ;) will fix

PR #278, opened 2019-03-05
Rename node config

dincho:


ansible/deploy.yml
```diff
@@ -98,7 +98,6 @@
           paths: "{{ packages_path }}"
           patterns:
             - .*aeternity-.*ubuntu.*.tar.gz
-            - .*epoch-.*ubuntu.*.tar.gz
```
valkov-ivaylo: Can this safely be removed? When this was added, the idea was that there  still might be some old downloads on machines. Files downloaded before epoch was renamed.

dincho: We don't deploy this name some time already, I doubt there is a leftovers, even it is it's probably a single file.

PR #290, opened 2019-03-14
fixing restore playbook

michalbagrowski:
fixing restore playbook, old restore mnesia required to have as much as memory as database is big.

### DIFF
--- a/ansible/mnesia_snapshot_restore.yml
+++ b/ansible/mnesia_snapshot_restore.yml
@@ -36,9 +36,9 @@
       tags: [health-check]
 
     - name: Restore Mnesia database
-      command: "{{ aeternity_bin }} eval '{atomic, _} = mnesia:restore(\"/tmp/backup\", [{default_op, recreate_tables}]).'"
+      command: "{{ aeternity_bin }} eval 'ok = mnesia:install_fallback(\"/tmp/backup\").'"
       register: restore_result
-      changed_when: "restore_result.stdout|search('atomic')"
+      changed_when: "restore_result.stdout|search('ok')"
 
     - import_tasks: tasks/restore_mining_and_peers_config.yml
 
--- a/ansible/templates/aeternity.backup.yaml
+++ b/ansible/templates/aeternity.backup.yaml
@@ -11,7 +11,7 @@ fork_management:
 mining:
     autostart: false
 
-
+peers: []
 keys:
   dir: keys
   peer_password: secret


ansible/mnesia_snapshot_restore.yml
```diff
@@ -36,9 +36,9 @@
       tags: [health-check]
 
     - name: Restore Mnesia database
-      command: "{{ aeternity_bin }} eval '{atomic, _} = mnesia:restore(\"/tmp/backup\", [{default_op, recreate_tables}]).'"
+      command: "{{ aeternity_bin }} eval 'ok = mnesia:install_fallback(\"/tmp/backup\").'"
```
dincho: I'm a bit confused. I see the default callback is [`mnesia_backup` in the docs](http://erlang.org/doc/man/mnesia.html#install_fallback-2)

ansible/mnesia_snapshot_restore.yml
```diff
@@ -36,9 +36,9 @@
       tags: [health-check]
 
     - name: Restore Mnesia database
-      command: "{{ aeternity_bin }} eval '{atomic, _} = mnesia:restore(\"/tmp/backup\", [{default_op, recreate_tables}]).'"
```
dincho: Also I think we also need to somehow pass `recreate_tables`, otherwise the backup's schema is not considered during the restore.
While this should be fine on restoring on new instances, it's a problem when the database is already there and have non-compatible schema.

As we use this only on new nodes, maybe track this for further improvement if not's trivial to pass the option in the PR.

ansible/mnesia_snapshot_restore.yml
```diff
@@ -36,9 +36,9 @@
...
+      command: "{{ aeternity_bin }} eval 'ok = mnesia:install_fallback(\"/tmp/backup\").'"
       register: restore_result
-      changed_when: "restore_result.stdout|search('atomic')"
+      changed_when: "restore_result.stdout|search('ok')"
```
dincho: isn't exact match more reliable ?

uwiger:
As discussed, a faster, more scalable, and conservative alternative might be to:
- stop the node
- perform a file-based backup of the database directory

There are tricks one could pull to force mnesia into a temporarily consistent state (e.g. pausing transaction log dumps) and do a quick snapshot while running, but that seems like overkill for now.

sennui:
Another alternative is to use checkpoints. (activate checkpoint -> backup checkpoint -> backup)
It may also ease issues with memory

michalbagrowski:
closing changes are in https://github.com/aeternity/infrastructure/pull/297

PR #291, opened 2019-03-14
fix subnet configuration

michalbagrowski:
fixing our deployment be equal across avalabality zones.

This is destructive change, see CircleCi task for details


michalbagrowski:
@dincho this change is rotating all static nodes :/

dincho:
@michalbagrowski maybe we should wait until we have proper chain restore in place ?
Is this blocker for some task?

michalbagrowski:
@dincho p.s. this is required by mainnet api gateway. because of issue with creating loadbalancer when subnets are in same avalabality zone

dincho:
So, we merge backup/restore, then this one, then api gateway ?

PR #295, opened 2019-03-18
Add bash host health check script pt 164331074

valerifilipov:


test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
...
+check_top_height() {
+    test $(curl -sS -m5 http://$HOST:3013/v2/key-blocks/current | jq '.height') -gt $height
+}
+check_version() {
+    test $(echo $node_status| jq -r '.node_version') == $version
+}
+
+node_status=$(get_node_status)
+
+check_genesis_hash || (echo "genesis check failed"; exit 1)
```
dincho: It would be nice to see all errors at once

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#example usage: test/scripts/health_check.sh --network=main --height=52000 --version=2.1.0 18.136.37.63
+set -eo pipefail
+
+main_network_id=ae_mainnet
```
dincho: isn't a hashmap more suitable and readable ?

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#example usage: test/scripts/health_check.sh --network=main --height=52000 --version=2.1.0 18.136.37.63
```
dincho: `height` should be really a `min_height`

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
...
+    curl -sS -m5 http://$HOST:3013/v2/status
+}
+check_genesis_hash() {
+    test $(echo $node_status| jq -r '.genesis_key_block_hash') == ${!network_genesis}
+}
+check_network_id() {
+    test $(echo $node_status| jq -r '.network_id') ==  ${!network_id}
+}
+check_top_height() {
+    test $(curl -sS -m5 http://$HOST:3013/v2/key-blocks/current | jq '.height') -gt $height
```
dincho: perhaps `-gte` ?

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#example usage: test/scripts/health_check.sh --network=main --height=52000 --version=2.1.0 18.136.37.63
+set -eo pipefail
+
+main_network_id=ae_mainnet
+uat_network_id=ae_uat
+main_genesis="kh_pbtwgLrNu23k9PA6XCZnUbtsvEFeQGgavY4FS2do3QP8kcp2z"
+uat_genesis="kh_wUCideEB8aDtUaiHCtKcfywU6oHZW6gnyci8Mw6S1RSTCnCRu"
```
dincho: We can hardcode main and uat minimum hight as well

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
...
+done
+
+if [ $# -eq 0 ]; then
+    HOST=127.0.0.1
+else
+    HOST=${@:$#}
+fi
+
+network=${network:-main}
+height=${height:-0}
```
dincho: if you change to `gte`, the default should be `1`

test/scripts/health_check.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#example usage: test/scripts/health_check.sh --network=main --height=52000 --version=2.1.0 18.136.37.63
+set -eo pipefail
+
+main_network_id=ae_mainnet
+uat_network_id=ae_uat
+main_genesis="kh_pbtwgLrNu23k9PA6XCZnUbtsvEFeQGgavY4FS2do3QP8kcp2z"
+uat_genesis="kh_wUCideEB8aDtUaiHCtKcfywU6oHZW6gnyci8Mw6S1RSTCnCRu"
+node_version=2.1.0
```
dincho: I don't think we should have version default

test/scripts/health_check.sh
valerifilipov: what number ?

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM alpine:3.8
 # Some of the playbooks run on multiple hosts: local + remote.
 # If it's set in the inventory it will not work when a specific inventory is used.
 # OpenSSL required for a packer workaround: https://github.com/hashicorp/packer/issues/2526
-RUN apk add --no-cache bash curl unzip make python3 py-cryptography openssh-client openssl sshpass jq\
+RUN apk add --no-cache bash curl unzip make python3 py-cryptography openssh-client openssl sshpass jq bc\
```
dincho: do we need `bc` in the setup.yml ?

test/scripts/health_check.sh
```diff
@@ -1,26 +1,41 @@
 #!/bin/bash
```
dincho: I'm not sure we'll even ship the `test` directory, but I think we want this script as part of the package, thus better path would be `scripts` dir in the root repo ?

Dockerfile
valerifilipov: yes, i missed that

scripts/health_check.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+#example usage: health_check.sh --network=main --min_height=52000 --version=2.1.0 18.136.37.63
+set -eo pipefail
+
+declare -A network_genesis_hashmap
```
dincho: This is bash 4, OSX still ships bash 3

I wonder should we support bash 3 and OSX. @michalbagrowski what's your opinion ?

scripts/health_check.sh
```diff
@@ -0,0 +1,96 @@
...
+network_genesis_hashmap['main']="kh_pbtwgLrNu23k9PA6XCZnUbtsvEFeQGgavY4FS2do3QP8kcp2z"
+network_genesis_hashmap['uat']="kh_wUCideEB8aDtUaiHCtKcfywU6oHZW6gnyci8Mw6S1RSTCnCRu"
```
dincho: maybe also hardcode min_height for those networks ?

Main at `54500` and uat at `54400`, @lucafavatella what's your opinion ?

scripts/health_check.sh
michalbagrowski:
```↪ ./scripts/health_check.sh --network=main --min_height=54000 --version=2.2.0 18.136.37.63
./scripts/health_check.sh: line 5: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]```

PR #297, opened 2019-03-19
Change backup restore method

michalbagrowski:


valerifilipov:
Are we going ti run it against multiple hosts? https://github.com/aeternity/infrastructure/pull/297/files#diff-f03e3fffcda2434f4599c932af720264L5

Makefile
```diff
@@ -69,6 +69,12 @@ reset-net: check-deploy-env
 		reset-net.yml
 
 mnesia_snapshot:
+ifeq ($(BACKUP_DB_VERSION),)
```
dincho: maybe we should combine BACKUP_DB_VERSION and DEPLOY_DB_VERSION in some more generic name and use in both cases

michalbagrowski: @dincho sure propose generic name i will change it...

dincho: Just a suggestion, not for this PR

michalbagrowski: Ok gotcha

scripts/bootstrap.sh
```diff
@@ -117,3 +121,16 @@ ansible-playbook \
...
     -e db_version=1 \
     deploy.yml
+
+if [ "$RESTORE_DATABASE" = true ] ; then
+    if [ "$env" = "main" ] || [ "$env" == "uat" ] ; then # restore only main / uat
+        ansible-playbook \
+            -i /tmp/local_inventory \
+            -e ansible_python_interpreter=$(which python3) \
+            --become-user aeternity -b \
+            -e download_dir=/tmp \
```
dincho: do we still need this?

michalbagrowski: no

PR #300, opened 2019-03-21
add job name in integration tests env name

valerifilipov:
https://www.pivotaltracker.com/story/show/164797055

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,11 +80,11 @@ commands:
       - run:
           name: Integration Tests
           command: |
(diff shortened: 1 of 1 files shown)


michalbagrowski:
are we sure that env name will not exeed some aws naming limit? it used in many places with long job name we might hit some limitsx

dincho:
I see the [limit is 256](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html)

@valerifilipov please calculate estimated length prefix+sha1+job name

valerifilipov:
I couldn't find limits for all resources but for naming and tags it's generally 255. We are going to be somewhere at half

PR #301, opened 2019-03-21
Goss local checks

michalbagrowski:
pt:  https://www.pivotaltracker.com/story/show/164334010

this is init commit, @dincho please review and correct me if i'm on wrong path.


ansible/setup.yml
```diff
@@ -132,6 +133,29 @@
...
+        - goss-install
+
+    - name: Install Goss binary
+      copy:
+        src: /usr/src/goss-{{ goss_version }}/goss
+        dest: /usr/bin/goss
+        mode: 0755
+        remote_src: true
+      tags:
+        - vault-install
```
valerifilipov: wrong tag ?

test/goss/local/health.yml
```diff
@@ -0,0 +1,32 @@
...
+    running: true
+
+port:
+  tcp:22:
+    listening: true
+    ip:
+    - 0.0.0.0
+
+port:
+  tcp:3015:
```
valerifilipov: i think these three `tcp:` should be under single `port:` because that way only the last will be parsed

PR #310, opened 2019-03-26
Add next environment seed peers

dincho:


test/goss/remote/vars/seed-peers-unstable.yaml
```diff
@@ -1,3 +0,0 @@
-hosts:
```
lucafavatella: Removal of ansible/group_vars/tag_env_unstable/vars.yml ?

dincho: Thanks for noticing, removed in 41c7f9e

PR #318, opened 2019-03-27
add network_id based on ansible vars file

valerifilipov:
Regarding https://www.pivotaltracker.com/story/show/164864896

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -44,6 +44,12 @@
       register: instance_tags
       when: is_aws
 
+    - name: "Load a variable files for environment: {{ env }}"
+      include_vars: "{{ item }}"
+      with_first_found:
+        - "vars/aeternity/{{ env }}.yml"
+        - "vars/aeternity/default.yml"
+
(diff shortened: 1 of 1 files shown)


ansible/monitoring.yml
```diff
@@ -83,6 +89,7 @@
         - "flavor:c{{ ansible_processor_vcpus }}.m{{ (ansible_memtotal_mb/1024)|round|int }}"
         - "public_ipv4:{{ public_ipv4 }}"
         - "role:{{ node_role|default('unknown') }}"
+        - "network_id:{{ node_config.fork_management.network_id|default('unknown') }}"
```
dincho: I think this will fail if there is no fork_management defined

michalbagrowski: good catch, i will have same issue with backup template file.

valerifilipov:
Fails with Mandatory variable ''db_version'' not defined

ansible/monitoring.yml
valerifilipov: fixed in 7b0b3dd

PR #320, opened 2019-03-28
add remote health_check.sh to goss

michalbagrowski:


test/goss/local/health.yml
```diff
@@ -4,6 +4,10 @@ user:
   master:
     exists: true
 
+command:
+  bash /tmp/health_check.sh:
```
dincho: I don't like the hardcoded location to the /tmp.
This validator should be able to run without the ansible playbook as well. So perhaps default a relative .sh script with an option to pass its path as variable/parameter.

michalbagrowski:
closing because of: https://github.com/aeternity/infrastructure/pull/322

PR #321, opened 2019-04-01
close 3013 for public for api gateway 164955096

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/164955096

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -257,7 +257,6 @@
           host: "localhost"
           timeout: 300
         when: api_gateway
-        connection: local
         tags: [health-check]
 
       - name: Refresh aeternity facts
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -239,7 +239,7 @@
           port: "{{ node_config.http.external.port }}"
           host: "{{ public_ipv4 }}"
           timeout: 300
-        when: health_check and not api_gateway
+        when: health_check and not local_health_check
         connection: local
         tags: [health-check]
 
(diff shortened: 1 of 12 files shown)


ansible/group_vars/all/vars.yml
```diff
@@ -1,2 +1,3 @@
 datadog_enabled: no
+api_gateway: no
```
dincho: A better variable name would be "local_health_check". Also what about using vars/aeternity/* ? I hope we'll remove ansible group vars usage soon, it does not scale well for our usage.

ansible/deploy.yml
```diff
@@ -239,15 +239,24 @@
...
+        when: health_check and not api_gateway
         connection: local
         tags: [health-check]
 
       - name: API health check
         uri:
           url: "{{ api_base_uri }}/blocks/top"
           timeout: 30
-        when: health_check
+        when: health_check and not api_gateway
```
dincho: what if you set `health_check: false` for `api-main` instead of this additional check?

ansible/deploy.yml
```diff
@@ -239,15 +239,24 @@
...
+        connection: local
+        tags: [health-check]
+
+      - name: Wait aeternity node API to boot gateway
+        wait_for:
+          port: "{{ node_config.http.external.port }}"
+          host: "localhost"
+          timeout: 300
+        when: api_gateway
         connection: local
```
dincho: Are you sure `connection: local` works with host:localhost ? I think that this task will be run locally on the controller host and do a request to the controller host

ansible/deploy.yml
michalbagrowski: that was my orginal idea, but you told me to have additional variable.

dincho: I guess it was misunderstanding  👍

PR #325, opened 2019-04-03
03.04.2019 stable release

michalbagrowski:
for resolving: https://www.pivotaltracker.com/story/show/164893447


ansible/files/goss/goss.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Goss
+
+[Service]
```
valkov-ivaylo: I'll suggest adding `Restart=on-failure`. According to [systemd documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) the default is `no`. If there are issues related to restarting the process from systemd on failure we might need some check (datadog ?) if goss service is not running.

dincho: This is just sync PR. This feature has been introduced in https://github.com/aeternity/infrastructure/pull/322

Could you please create PT ticket to improve that, I do agree with the comment otherwise.

PR #327, opened 2019-04-03
add deploy region

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/162341749

@dincho please point me what should be added more in docs.


dincho:
I think region+rolling should be good enough. Could you please add ROLLING_UPDATE to https://github.com/aeternity/infrastructure/pull/327/files#diff-04c6e90faac2675aa89e2176d2eec7d8R185 as well ?

PR #328, opened 2019-04-03
fix security rules for gateway.

michalbagrowski:


michalbagrowski:
because of: https://circleci.com/gh/aeternity/infrastructure/10551

PR #349, opened 2019-04-11
Add release artifacts validation script pt 164978469

valerifilipov:
In scope of [PT-164978469](https://www.pivotaltracker.com/story/show/164978469)
Should be run in aeternity/aeternity CI on tag

### DIFF
--- a/scripts/check_release_artifacts.sh
+++ b/scripts/check_release_artifacts.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+# should be run in CI or manually as:
+# ./check_release_artifacts.sh v2.3.1
+set -e
+
+if [[ -n "$1" ]]; then
+    release=${1}
+else
+    echo specify release version e.g. ${0} v2.3.0 >&2; exit 1
+fi
+
+failed=false
+if [[ -n $release && $release =~ ^v([0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9]+)*)$ ]]; then
+    version=${BASH_REMATCH[1]}
+else
+    version=$release
+fi
+
+declare -a targets=(ubuntu macos)
+repo_name=aeternity/aeternity
+
+check_release_asset() {
+    curl -fs -o /dev/null --head $1|| echo $1 check failed; failed=true
+}
+
+check_github_release_assets() {
+    for target in "${targets[@]}"; do
+        GH_ASSET=https://github.com/$repo_name/releases/download/$release/aeternity-$version-$target-x86_64.tar.gz
+        check_release_asset $GH_ASSET
+        # TODO: compare MD5 checksum with local build
+    done
+}
+
+get_dockerhub_token() {
+    curl --silent --user "${DOCKER_USER}:${DOCKER_PASS}" \
+        "https://auth.docker.io/token?scope=repository:$repo_name:pull&service=registry.docker.io" \
+        |jq -r .token
+}
+
+get_docker_image_digest() {
+    local image=$1
+    local tag=$2
+    local token=$3
+   curl \
+    --silent \
+    --header "Accept: application/vnd.docker.distribution.manifest.v2+json" \
+    --header "Authorization: Bearer $token" \
+    "https://index.docker.io/v2/$image/manifests/$tag" \
+    | jq -r '.config.digest'
+}
+
+get_docker_image_config() {
+  local image=$1
+  local token=$2
+  local digest=$3
+
+  curl \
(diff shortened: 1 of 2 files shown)


scripts/check_release_artifacts.sh
```diff
@@ -0,0 +1,120 @@
...
+
+  curl \
+    --silent \
+    --location \
+    --header "Authorization: Bearer $token" \
+    "https://registry-1.docker.io/v2/$image/blobs/$digest" \
+    | jq -r '.container_config'
+}
+
+check_dockerhub_assets() {
```
dincho: but whyyy !?? ;(

```
$ curl https://index.docker.io/v1/repositories/aeternity/aeternity/tags/v2.0.0
[]
dincho at Dinchos-Mac-mini in ~/Projects/aeternity/infrastructure on health_checks [$]
$ curl https://index.docker.io/v1/repositories/aeternity/aeternity/tags/v2.0.5
Tag not found
```

https://stackoverflow.com/questions/32113330/check-if-imagetag-combination-already-exists-on-docker-hub/37799618

scripts/check_release_artifacts.sh
```diff
@@ -0,0 +1,120 @@
...
+
+check_s3_artifacts() {
+    for target in "${targets[@]}"; do
+        local version_package=https://releases.ops.aeternity.com/aeternity-$version-$target-x86_64.tar.gz
+        local latest_package=https://releases.ops.aeternity.com/aeternity-latest-$target-x86_64.tar.gz
+        # TODO: calculate ETAG from local package MD5 sum and verify
+        # https://teppen.io/2018/10/23/aws_s3_verify_etags/
+        local latest_etag=$(get_s3_package_etag $latest_package)
+        local version_etag=$(get_s3_package_etag $version_package)
+        if [[ "$version_etag" == "" ]]; then
```
dincho: could you please simplify this ?
It should really be only `"$version_etag" != "$latest_etag"`

scripts/check_release_artifacts.sh
```diff
@@ -0,0 +1,120 @@
...
+declare -a targets=(ubuntu macos)
+repo_name=aeternity/aeternity
+
+check_release_asset() {
+    curl -fsS -o /dev/null --head "$1" 2>&1
+}
+
+check_github_release_assets() {
+    for target in "${targets[@]}"; do
+        GH_ASSET=https://github.com/$repo_name/releases/download/$release/aeternity-$version-$target-x86_64.tar.gz
```
dincho: releases.ops.aeternity.com should be checked as well

scripts/check_release_artifacts.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+# should be run in CI or manually as:
+# ./check_release_artifacts.sh v2.3.1
```
dincho: could you please remove the `v` from the parameter, it's implementation/naming detail.

scripts/check_release_artifacts.sh
valerifilipov: This is not used, I'll remove it. It could be used if we want to compare remote image digest with local build image.  (did not dig into this much, just wanted to see what it prints)

scripts/check_release_artifacts.sh
valerifilipov: But that's what comes from CIRCLE_TAG, and could be used as here:
https://circleci.com/gh/aeternity/aeternity/22338
I'll post a pull request there later

dincho: 🤔 yeah, true. We could tag (and validate) v2.3.1-beta2 as well. 👍

scripts/check_release_artifacts.sh
valerifilipov: fixed in 3c5e3db

scripts/check_release_artifacts.sh
valerifilipov: fixed in 21b3251

scripts/check_release_artifacts.sh
valerifilipov: It's checked by `check_s3_artifacts`

scripts/check_release_artifacts.sh
dincho: I don't think you got the idea, this check can be done with single GET request, you don't need tokens and multiple REST API calls.

dincho: Nevermind, digest check is a good addition.

PR #355, opened 2019-04-12
stop goss service logging PT-165273736

valerifilipov:
Regarding [PT-165273736](https://www.pivotaltracker.com/story/show/165273736)


ansible/files/goss/goss.service
```diff
@@ -7,6 +7,7 @@ ExecStart=/usr/bin/goss -g /etc/goss/health.yml serve &
 ExecStop=/bin/kill -WINCH ${MAINPID}
 KillSignal=SIGKILL
 User=aeternity
+StandardOutput=null
```
dincho: what about init.d ? Should we care about it at all?

valerifilipov: It's not used since there is a service file. Also does not log to syslog and not managed by rsyslogd

PR #362, opened 2019-04-17
Use AWS tags to bootstrap instances

dincho:
The changes are user_data related, which was split in external repository: https://github.com/aeternity/infra-node-aws-bootstrap

I was hoping to be able to have a BC changes and the rotation to be applied later, however using TF variable in user_data configuration field also introduces rotation/change unfortunately.

So, this PR needs code/state freeze and manual application to all environments.


terraform/modules/cloud/aws/deploy/fleet/main.tf
```diff
@@ -125,7 +131,7 @@ resource "aws_launch_configuration" "spot-with-additional-storage" {
 }
 
 data "template_file" "spot_user_data" {
-  template = "${file("${path.module}/templates/spot_user_data.bash")}"
+  template = "${file("${path.module}/templates/${var.spot_user_data_file}")}"
```
valerifilipov: Why use two templates and two vars, when they point to the same file now?

terraform/modules/cloud/aws/deploy/fleet/templates/user_data.bash
```diff
@@ -1,16 +1,7 @@
...
-# set -x
-exec > >(tee /tmp/user-data.log|logger -t user-data ) 2>&1
+set -eo pipefail
 
-if [ ! -d "/infrastructure" ] ; then
-    git clone -b ${bootstrap_version} --single-branch https://github.com/aeternity/infrastructure.git /infrastructure
-else
-    git -C /infrastructure fetch origin ${bootstrap_version}
-    git -C /infrastructure reset --hard origin/${bootstrap_version}
-fi
```
valerifilipov: Where do ansible playbooks come from ?

dincho: https://github.com/aeternity/infrastructure/pull/362/files#diff-d3b479e0cb63584581a60025e37dfa0aR7

terraform/modules/cloud/aws/deploy/fleet/main.tf
dincho: In case we need difference user data as before.

PR #367, opened 2019-04-24
remove group_vars PT-164627692

valerifilipov:
In scope of [PT-164627692](https://www.pivotaltracker.com/story/show/164627692)

### DIFF
--- a/ansible/vars/aeternity/dev2.yml
+++ b/ansible/vars/aeternity/dev2.yml
@@ -1,3 +1,4 @@
+datadog_enabled: yes
 api_base_uri: http://{{ public_ipv4 }}:{{ node_config.http.external.port }}/v2
 configure_peers: true
 seed_peers: []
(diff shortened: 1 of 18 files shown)


ansible/vars/aeternity/dev2.yml
```diff
@@ -1,3 +1,5 @@
+public_ipv4: "{{ ansible_ec2_public_ipv4|default(ansible_ssh_host)|default(ansible_host)|default(inventory_hostname) }}"
+datadog_enabled: yes
```
dincho: It should be enabled on dev2

valerifilipov: There is no group_vars for dev2, so I disabled it

dincho: ehm, yes I meant it should be disabled :)

PR #369, opened 2019-04-24
shrink main and gw nodes root volumes to 8G

valerifilipov:
In scope of https://www.pivotaltracker.com/story/show/164978255


dincho:
This one needs manual application for some of the environments

dincho:
Already Irrelevant PR after the split.

PR #372, opened 2019-04-30
Use tags instead of stable infrastructure branch pt 165567448

valerifilipov:
re https://www.pivotaltracker.com/story/show/165567448

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,8 +187,9 @@ jobs:
           name: Tag and push Docker image to DockerHub on Git tags
           command: |
             docker login -u $DOCKER_USER -p $DOCKER_PASS
-            docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} .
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -268,7 +269,8 @@ workflows:
             - build_check
           filters:
             branches:
-              only: stable
+              ignore: /.*/
+              only: /^v.*$/
```
dincho: This has to be under tags.

ansible/async_provision.yml
```diff
@@ -5,7 +5,7 @@
   become: yes
 
   vars:
-    var_bootstrap_version: "{{ bootstrap_version | default('stable', true) }}"
+    var_bootstrap_version: "{{ bootstrap_version | default('v2.0.1, true) }}"
```
dincho: let's make `master` the default, or remove the default at all.

The hardcoded version here will only have us headaches

PR #373, opened 2019-05-07
Split terraform modules in separate repos pt 165295661

valerifilipov:
re https://www.pivotaltracker.com/story/show/165295661
needs git installed in master docker image


.dockerignore
```diff
@@ -8,7 +8,7 @@
 terraform/.terraform
 !Makefile
 !requirements.txt
-!test/goss/
```
dincho: What's the reason for this change ?

valerifilipov: `test/terraform` needs to be in docker for integration tests in https://github.com/aeternity/terraform-aws-aenode-deploy/pull/1 to run

terraform/environments/main.tf
```diff
@@ -3,7 +3,7 @@ variable "vault_addr" {
 }
 
 module "aws_deploy-main-ap-southeast-1" {
-  source            = "../modules/cloud/aws/deploy"
+  source            = "git::https://github.com/aeternity/terraform-aws-aenode-deploy.git"
```
lucafavatella: How is versioning handled? 🤔

lucafavatella: I mean - commit in this repo infrastructure shall link to immutable reference to the terraform-aws-aenode-deploy repo. And I see no version / tag / commit hash of terraform-aws-aenode-deploy in this line.

dincho: It defaults to HEAD, @valerifilipov perhaps we should tag module v1.0.0 (as it's production tested already) and reference it here ?

dincho: Also why not prefer the short version `github.com/aeternity/terraform-aws-aenode-deploy` ?

valerifilipov: Yes, we should tag both repos, after https://github.com/aeternity/terraform-aws-aenode-deploy/pull/1 and https://github.com/aeternity/terraform-aws-api-gateway/pull/1 get merged

dincho: You don't need CI integrations to tag it

.dockerignore
valerifilipov: removed as actually not used

terraform/environments/main.tf
```diff
@@ -3,7 +3,7 @@ variable "vault_addr" {
 }
 
 module "aws_deploy-main-ap-southeast-1" {
-  source            = "../modules/cloud/aws/deploy"
+  source            = "github.com/aeternity/terraform-aws-aenode-deploy?ref=v1.0.0"
```
lucafavatella: I see there is https://github.com/aeternity/terraform-aws-aenode-deploy/tree/v1.0.0

Not sure about syntax though it makes sense.

PR #377, opened 2019-05-09
Add GITHUB_API_TOKEN to vault

dincho:
This should allow us to remove it from CI environment vars config


import-secrets.sh
```diff
@@ -40,6 +40,8 @@ if [ -n "$VAULT_TOKEN" ]; then
     export GOOGLE_APPLICATION_CREDENTIALS=$HOME/.gcp.json
     vault read -field json secret/google/${VAULT_SECRETS_ROLE} > $GOOGLE_APPLICATION_CREDENTIALS
 
+    export GITHUB_API_TOKEN=$(vault read -field=value secret/github/prod/token)
```
lucafavatella: I see https://github.com/aeternity/aeternity/blob/cd9bbb63baed007e94921df7c7c92b55a5a0efda/.circleci/config.yml#L251

PR #381, opened 2019-05-15
Improve secrets import

dincho:
https://www.pivotaltracker.com/story/show/165938339

Rework the secrets authentication and usage scheme.
Main benefits:
- role/policy based secrets
- import/dump with make recipe which has few benefits
  - ensure it's dumped only once (currently in CI it's imported multiple time when a new bash script is run, because of CirlceCI mechanics)
  - secrets prerequisites so that it's handled automatically when running make targets (i.e. docker entry-point is not needed anymore)
  - easies Vault token (and all secrets) refresh
- more explicit secrets usage (CI config, makefile)
- dumping to FS and using `envdir` is compatible with [Vault agent](https://www.vaultproject.io/docs/agent/) which we can use in future instead of custom scripts (hopefully 🤞 

Cons:
- additional `envdir` dependancy
- secrets are stored in the FS, however given that we use only ephemeral environments (docker container console, CI VMs/containers) this shouldn't be an issue.

The old `import-secrets.sh` is kept for backward compatibility until dependant CI configurations are migrated


scripts/secrets/policies/node-ci.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+VAULT_SECRETS_ROLE=${VAULT_SECRETS_ROLE:-ae-inventory}
+
+source ${DIR}/policies/ci.sh
+
+ROCKET_HOOK_URL=$(vault read -field=core-alerts secret/rocketchat/prod/hooks)
+
+dump_var "ROCKET_HOOK_URL"
```
lucafavatella: Based on this `dump_var` not defined in this file, and seeing that intended usage of this file seems to source it, it seems to me that `#!/bin/bash` shebang is misleading. Shall not `#!/bin/bash` shebang be deleted from files not meant to be executed? I also wonder if a conventional extension exists for these case.

dincho: I was in doubt when adding them as well. I removed it in 79eb61b

PR #383, opened 2019-05-16
Disable hourly seed checks and daily deploy in CI

valerifilipov:
[PT-166025614](https://www.pivotaltracker.com/story/show/166025614)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,28 +334,3 @@ workflows:
           context: ae-infra-manage
       - chain_snapshot_main:
           context: ae-infra-manage
-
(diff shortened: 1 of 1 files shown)


lucafavatella:
Will these be moved to other location for scheduled execution?

dincho:
> Will these be moved to other location for scheduled execution?

I never liked the hourly seed checks in the CI, so we should looks fo better alternatives than adding it back to CI. It should be somehow integrated with the monitoring.

In regards to daily deploy it was playing a role of integration tests in the past. As we now have integration tests with relatively good coverage, I don't think we need that anymore.

PR #391, opened 2019-05-28
Cleanup

dincho:
https://www.pivotaltracker.com/story/show/166268239

Most of the reasons can be found in the commit messages.


ansible/tasks/migrate-db-name.yml
```diff
@@ -1,44 +0,0 @@
...
-- name: Stop aeternity node
-  command: "{{ aeternity_bin }} stop"
-  when:
-    - binary.stat.exists
-    - ping.stdout == "pong"
-    - db.stat.exists
-    - find_results.files|length == 0
-
-- name: Migrate db schema names
-  command: "{{ aeternity_bin }} rename_db {{ db_path }}"
```
lucafavatella: Shall we remove in the next major release [the script for renaming](https://github.com/aeternity/aeternity/blob/a6ab0c6dd7b9b25ae1ce8af271bbdf98cd2cf52a/rebar.config#L150-L163.)

Current node release supports [2.2.0+](https://github.com/aeternity/aeternity/blame/a6ab0c6dd7b9b25ae1ce8af271bbdf98cd2cf52a/docs/release-notes/RELEASE-NOTES-3.0.0-rc.1.md#L65) and rename was in [2.2.0](https://github.com/aeternity/aeternity/blame/a6ab0c6dd7b9b25ae1ce8af271bbdf98cd2cf52a/docs/release-notes/RELEASE-NOTES-2.2.0.md#L10).

ansible/deploy.yml
```diff
@@ -140,16 +139,6 @@
           mode: 0700
         tags: [config, peer_keys]
 
-      - name: "Copy node keys to {{ node_config['keys']['dir'] }}"
-        copy:
-          src: "{{ item }}"
-          dest: "{{ project_root }}/{{ node_config['keys']['dir'] }}"
```
lucafavatella: I need to understand if this is sync/gossip peer keys for seed nodes.

ansible/tasks/migrate-db-name.yml
dincho: This repository while public is intended for internal usage, and we don't have any 2.2.0 already.

In regards to the node script, yes it makes sense to remove it in next major.

ansible/deploy.yml
dincho: I can't really find in the history, but this was the mining keys used to sign the blocks. Probably even before enoise

ansible/deploy.yml
```diff
@@ -182,32 +149,7 @@
...
-        copy:
-          content: "{{ genesis_accounts | to_nice_json }}"
-          dest: "{{ genesis_accounts_path }}"
-        when: genesis_accounts is defined
-        tags: [config]
-
-      - name: "Copy node public peer key to {{ project_root }}/{{ node_config['keys']['dir'] }}/peer_key.pub"
+      - name: "Copy node public peer key to {{ node_config['keys']['dir'] }}/peer_key.pub"
         copy:
           content: "{{ lookup('hashi_vault', 'secret=secret/aenode/peer_keys/{{ public_ipv4 }}/public:base64')|b64decode }}"
```
lucafavatella: I see peer keypair taken from vault.

dincho:
> I am convinced that the on deploy the peer keypair is taken from vault so ok.

Yes

dincho:
> I need to understand reset-net.

I didn't looked into it at all

PR #394, opened 2019-06-04
Dont' require configuration file to exist in addional storage playbook

dincho:
Should fix https://www.pivotaltracker.com/story/show/166438273

This set of tasks depends on `additional_storage` to execute. Without env configuration file it should not run.

### DIFF
--- a/ansible/tasks/additional_storage.yml
+++ b/ansible/tasks/additional_storage.yml
@@ -15,7 +15,9 @@
 - name: "Load a variable files for environment: {{ env }}"
   include_vars: "{{ item }}"
   with_first_found:
-    - "vars/aeternity/{{ env }}.yml"
+    - files:
+        - "vars/aeternity/{{ env }}.yml"
+      skip: true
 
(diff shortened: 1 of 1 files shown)


ansible/tasks/additional_storage.yml
```diff
@@ -15,7 +15,9 @@
 - name: "Load a variable files for environment: {{ env }}"
   include_vars: "{{ item }}"
   with_first_found:
-    - "vars/aeternity/{{ env }}.yml"
+    - files:
+        - "vars/aeternity/{{ env }}.yml"
+      skip: true
```
lucafavatella: Does this skip only this step ("Load a variable files for environment ...") or also the rest of the file?

venimus: From the error message `fatal: [default]: FAILED! => {"msg": "No file was found when using with_first_found. Use the 'skip: true' option to allow this task to be skipped if no files are found"}`
This just skips this step and continue with the other tasks and try to resolve the variables by other means

PR #395, opened 2019-06-04
Increase SSH idle timeout

dincho:
https://www.pivotaltracker.com/story/show/166176328


docker/ssh_config
```diff
@@ -2,3 +2,5 @@ Host *
     User aeternity
     IdentityFile ~/.ssh/id_ae_infra_ed25519
     CertificateFile ~/.ssh/id_ae_infra_ed25519-%r-cert.pub
+    ServerAliveInterval 120
```
venimus: I think this is an option for the sshd_config (ssh server) while for the ssh_config (ssh client) it should be `ClientAliveInterval`

PR #398, opened 2019-06-10
Version lock of PIP tools

venimus:
Datadog 0.29.0 has a bug (https://github.com/DataDog/datadogpy/issues/390) that breaks the build so we need to lock version to 0.28 and for the other tools as well
In addition it appears that there is an issue with https://github.com/DataDog/ansible-datadog-callback/tree/2.4.2

https://www.pivotaltracker.com/story/show/166571685


Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN curl -L https://github.com/aelsabbahy/goss/releases/download/${GOSS_VER}/gos
 ADD requirements.txt /infrastructure/
 RUN apk add --no-cache --virtual build-deps \
         gcc python3-dev musl-dev openssl-dev libffi-dev linux-headers \
-    && pip3 install --upgrade pip==18.0 \
+    && pip3 install --upgrade pip==18.1 \
```
venimus: this is to avoid downgrading of the preinstalled `18.1` (saves few seconds)

PR #401, opened 2019-06-19
Fix missing db_path var for api_uat env

venimus:
Ad missing db_path var in `api_uat.yaml` env
and Check for missing `db_path` var if `persist: true` to prevent further misconfigurations

Node is deployed but ansible fails to create the db directory (as the var is undefined)

in scope of https://www.pivotaltracker.com/story/show/164978300

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -131,7 +131,10 @@
       - name: Fail on missing 'node_config.chain.db_path' variable
         fail:
           msg: 'Chain persistence is enabled, but "node_config.chain.db_path" is not defined'
-        when: node_config.chain.persist is defined and node_config.chain.persist and node_config.chain.db_path is not defined
+        when:
+          - node_config.chain.persist is defined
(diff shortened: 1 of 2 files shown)


ansible/deploy.yml
```diff
@@ -128,6 +128,12 @@
         notify: "restart aeternity daemon"
         tags: [package]
 
+      - name: Fail on missing 'node_config.chain.db_path' variable
+        fail:
+          msg: 'Chain persistence is enabled, but "node_config.chain.db_path" is not defined'
+        when: node_config.chain.persist is defined and node_config.chain.persist and node_config.chain.db_path is not defined
```
dincho: could you please split this in multiple line for readability ? (when takes a list)

venimus: https://github.com/aeternity/infrastructure/pull/401/commits/eee3fdbf274424c0a165828060f3b4a1c8be2bd3

PR #406, opened 2019-07-08
Update Terraform to latest stable

dincho:
While there are some BC breaking changes, the configuration migration should be harmless, I already updated the aenode module and api gateway module locally.


venimus:
The only issue i noticed is that we need to prefix inclusion of 'local' modules with `./`

PR #408, opened 2019-07-19
Export region in playbook

venimus:
Allow use of different publisher accounts per monitoring node (1 node per region). Keys are stored in the vault grouped by region

Currently we are using single account with multiple nodes and this causes conflicting or confusing stats data.

Biggest problem is that nonces will be racing all the time and this will produce a lot of nonce errors (and tx rejects). Each node will try to increase the nonce +1 but their starting nonce will be the same, thus 2 nodes will try to post a transaction with the same nonce, one will be successful but the rest will fail (and produce nonce error in the logs). In time nonce might eventually sync, but the problem may occur again under certain conditions.


venimus:
Build fails because the currently used terraform 0.12 requires locally referenced modules to be prefixed with `./` while is currently the dependent module 'aenode-deploy' is not updated. A refactoring [PR](https://github.com/aeternity/terraform-aws-aenode-deploy/pull/10) is pending. Once the module is merged it should pass

PR #410, opened 2019-08-02
Fix snapshot restore

dincho:
In scope of https://www.pivotaltracker.com/story/show/167595374

This is _soft_ backward compatibility break, that is, an environment that lack `snapshot_filename` tag will not be restored from a snapshot, however the environment should provision and sync just fine.

### DIFF
--- a/ansible/mnesia_snapshot.yml
+++ b/ansible/mnesia_snapshot.yml
@@ -1,9 +1,14 @@
 ---
-# Playbook used create mnesia snapshoot and upload it to S3
+# Playbook used create mnesia snapshot and upload it to S3
 
 - name: Backup, archive and download Mnesia database
   hosts: "{{ ansible_play_hosts | last }}"
   remote_user: aeternity
+
+  vars:
+    snapshot_suffix: snapshot
+    snapshot_filename: "mnesia_{{ env }}_v-{{ db_version }}_{{ ansible_ssh_host|default(inventory_hostname) }}_{{ snapshot_suffix }}.tgz"
+
   tasks:
     - name: Include configuration variables
       include_vars: vars/snapshot.yml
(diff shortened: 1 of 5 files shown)


ansible/mnesia_snapshot.yml
```diff
@@ -21,27 +26,35 @@
...
+
+    - name: Archive Mnesia Directory (with additional storage)
       archive:
-        format: "{{ format }}"
-        path: "{{ project_root + '/' if additional_storage is not defined else '' }}{{ node_config.chain.db_path }}"
-        dest: /tmp/{{ backup_archive_filename }}
+        format: gz
+        path: "{{ project_root }}/{{ node_config.chain.db_path }}"
+        dest: /tmp/{{ snapshot_filename }}
+      when: additional_storage is defined
```
venimus: and additional_storage

ansible/mnesia_snapshot.yml
```diff
@@ -21,27 +26,35 @@
 
     - meta: flush_handlers
 
-    - name: Archive Mnesia Directory
+    - name: Archive Mnesia Directory (no additional storage)
+      archive:
+        format: gz
+        path: "{{ project_root }}/{{ node_config.chain.db_path }}"
+        dest: /tmp/{{ snapshot_filename }}
+      when: additional_storage is not defined
```
venimus: or not addtional_storage

ansible/mnesia_snapshot.yml
dincho: fixed in 9c0cf35

ansible/mnesia_snapshot.yml
dincho: fixed in 9c0cf35

ansible/mnesia_snapshot.yml
```diff
@@ -21,39 +27,51 @@
...
         mode: put
         aws_access_key: "{{ lookup('env','AWS_ACCESS_KEY_ID') }}"
         aws_secret_key: "{{ lookup('env','AWS_SECRET_ACCESS_KEY') }}"
         security_token: "{{ lookup('env','AWS_SESSION_TOKEN') }}"
 
-    - name: Remove Mnesia backups
+    - name: Delete local Mnesia snapshots ({{ snapshots_dir }})
       file:
-        path:  "{{ item }}"
+        path: "{{ snapshots_dir }}"
```
lucafavatella: All? or snapshot_path?

dincho: Yes, it is meant to delete all snapshots in that directory. The use-case is if something fail in the middle, next time it's cleanedup.

PR #415, opened 2019-08-26
Enable internal API listener on mainnet API gateway nodes

dincho:
In scope of https://www.pivotaltracker.com/story/show/168074857


dincho:
> 😨

It's not exposed by default (firewall) to the public but only to the load-balancer, then only dry-run will be routed through it to the nodes

PR #417, opened 2019-09-12
Script to execute XFS grow of nodes

venimus:
After increasing EBS volumes it is required to grow the file system size
in scope of https://www.pivotaltracker.com/story/show/168429919

Note: only volume increase is allowed on-the-fly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,17 @@ endif
 		-e env=$(BACKUP_ENV) \
 		mnesia_snapshot.yml
 
+ebs-grow-volume: check-deploy-env secrets
+	$(eval LIMIT=tag_role_aenode:&tag_env_$(DEPLOY_ENV))
+ifneq ($(DEPLOY_REGION),)
+	$(eval LIMIT=$(LIMIT):&region_$(DEPLOY_REGION))
+endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -93,6 +93,24 @@ endif
 		-e env=$(BACKUP_ENV) \
 		mnesia_snapshot.yml
 
+%.yml: check-deploy-env secrets
```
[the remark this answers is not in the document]
venimus: it was actually a proposal. $PARAMS is to be able to add any additional params for ansible, like more env (-e) or inventory filter (-i).

ansible/ebs-grow-volume.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: Update FS size of EBS volume storage
+  hosts: all
+  remote_user: master
+  become: yes
+
+  tasks:
+    - name: Check if running on AWS instance
+      uri:
+        url: http://169.254.169.254/latest/meta-data
```
tolbrino: Is this a magic internal IP?

dincho: Yes, AWS meta service

PR #418, opened 2019-09-17
Add windows artifacts check

venimus:
Should be merged after artifacts are built properly

Note: Windows artifacts are .zip

### DIFF
--- a/scripts/check_release_artifacts.sh
+++ b/scripts/check_release_artifacts.sh
@@ -17,7 +17,7 @@ else
     version=$release
 fi
 
-declare -a targets=(ubuntu macos)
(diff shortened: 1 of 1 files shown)


dincho:
> Should be merged after artifacts are built properly

not necessary, the check is using the [stable image](https://github.com/aeternity/aeternity/blob/master/.circleci/config.yml#L1071) so this changes will not kick in in the aeternity CI until this repo is tagged

venimus:
Yes, but it might get tagged for some other reason, before the other PR is merged

dincho:
> Yes, but it might get tagged for some other reason, before the other PR is merged

Shouldn't be a problem if you're careful when tagging :)

venimus:
Accidentally pushed something from other branch

PR #456, opened 2019-12-04
Enable testnet-api HTTP cache

IvayloPapazov:
ref: https://github.com/aeternity/terraform-aws-testnet-api/issues/22


dincho:
Is the Vault config already updated as well ?

PR #461, opened 2019-12-18
Deploy makefile recipe should automatically create SSH cert

### LINKED ISSUE #433: Deploy makefile recipe should automatically create SSH cert
First the node CI should be migrated to dynamic certs instead of static keys

IvayloPapazov:
close #433


venimus:
why no `secrets` recipe require `cert`?

PR #479, opened 2020-01-30
Make mnesia_snapshot fails with missing config

### LINKED ISSUE #474: make mnesia_snapshot fails with missing config
Running `BACKUP_ENV=dev1 make mnesia_snapshot` fails with 
```
ERROR! Unable to retrieve file contents
Could not find or access '/tmp/config/dev1.yml' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
make: *** [Makefile:50: ansible/mnesia_snapshot.yml] Error 1
```

IvayloPapazov:
fix #474

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,8 @@ PYTHON ?= /usr/bin/python3
 DEPLOY_DOWNTIME ?= 0
 ROLLING_UPDATE ?= 100%
 BACKUP_SUFFIX ?= backup
-DEPLOY_ENV ?=
+BACKUP_ENV ?=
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -46,7 +46,9 @@ ANSIBLE_EXTRA_VARS =
 ANSIBLE_EXTRA_PARAMS ?=
 
 .PRECIOUS: $(DEPLOY_CONFIG)
-ansible/%.yml: cert | $(DEPLOY_CONFIG)
```
venimus: it should work if you remove only '|'

PR #488, opened 2020-03-19
Remove virtual env variables from Makefile

IvayloPapazov:


dincho:
While this is to be removed in future indeed, most of our nodes are still running 16.04 with venv, so let's keep this as default and use explicit python path for new 18.04 nodes, e.g.:
`PYTHON=/usr/bin/python3 make deploy`

PR #498, opened 2020-04-10
Disable root user login on all nodes

### LINKED ISSUE #489: Disable root user login on all nodes (ansible playbook)

IvayloPapazov:
close #489


IvayloPapazov:
> I guess that would disable any "debugging" access?

No. This is just a ssh restriction on root. Everything works just like it used to.

PR #499, opened 2020-04-27
Get aeternity package link from vault config

IvayloPapazov:


venimus:
I guess the requirement to ignore the package specified in the tag?

scripts/bootstrap.sh
```diff
@@ -89,7 +88,7 @@ ansible-playbook \
     --become-user aeternity -b \
     -e env=${env} \
     -e db_version=1 \
-    -e package=${aeternity_package} \
+    -e package=$(grep 'package:' /tmp/node_config.yml | awk '{ print $2 }') \
```
venimus, line 91: may be you could reuse the old variable to precalculate this (would be easier to maintain some time in the future)

dincho: This line is obsolete anyway as the `package` should be already in the /tmp/node_config.yml

venimus: yes, i recall the previous implementation was to take the aws tag as priority, so seems it is no longer needed if tag is deprecated

PR #501, opened 2020-05-28
Add mainnet sync phase to release process

IvayloPapazov:
Part of: https://github.com/aeternity/aeternity/issues/3213


ansible/integration_sync.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Delete Mnesia database
```
venimus, line 2: is this the proper title?

PR #503, opened 2020-06-12
When resetting the mnesia database, show an error if there are no synchronized nodes

IvayloPapazov:


ansible/mnesia_reset_once.yml
```diff
@@ -6,8 +6,30 @@
...
+          uri:
+            url: "http://{{ item }}:{{ node_config.http.external.port | default('3013') }}/v2/status"
+          register: hosts_status
+          loop: "{{ play_hosts }}"
+
+        - name: Check that the node is synchronized
+          set_fact:
+            missing_synced_node: false
+          when: item.json.sync_progress | int == 100
+          loop: "{{ hosts_status.results }}"
```
venimus, line 24: Not sure about the purpose and requirements, but it looks to me like that this will "fail" if 1 of the nodes is not synced, not all of them

ansible/mnesia_reset_once.yml
```diff
@@ -6,8 +6,30 @@
   vars:
     project_root: "{{ ansible_env.HOME }}/node"
     aeternity_bin: "{{ project_root }}/bin/aeternity"
+    missing_synced_node: true
```
venimus, line 9: this is kind of misleading imho. probably should be reversed like "have_synced_node: false" and later reverse the bool values

IvayloPapazov: 81d4ea7

ansible/mnesia_reset_once.yml
```diff
@@ -6,8 +6,30 @@
...
+        - name: Check that the node is synchronized
+          set_fact:
+            found_synced_node: true
+          when: item.json.sync_progress | int == 100
+          loop: "{{ hosts_status.results }}"
+
+        - name: Fail if no synced node found
+          fail:
+            msg: No synced node
+          when: found_synced_node == false
```
venimus, line 29: why no `when: not found_synced_node`

dincho:
> cant really get the trick here (to have the loop collect all node statuses) is this "delegate_to" doing it?

Yes, and it has `loop` and it's executed locally only.

PR #505, opened 2020-06-25
Send datadog event when resetting the mnesia database

IvayloPapazov:


ansible/mnesia_reset_once.yml
```diff
@@ -6,26 +6,59 @@
...
+          | selectattr('sync_progress', 'equalto', 100)
+          | list
+          | length == 0
+
+      - name: Send Datadog event
+        datadog_event:
+          host: "{{ public_ipv4 }}"
+          alert_type: success
+          title: Reset mnesia database
+          text: "Synced node found"
```
dincho, line 38: This is not very useful, I'd rather move it below after the actual node reset. Btw I just noticed we missed the usual node health checks.

In general one success notification should be sent once the full playbook completes (with additional health checks).
A failing notification should be sent regardless of the error/fail reason (of any task).
Nice to have is the actual failing task and reason carried in `ansible_failed_task` and `ansible_failed_result`
 (see the bottom of https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html)

PR #511, opened 2021-03-12
Fix deploy scripts

dincho:


scripts/deploy.sh
```diff
@@ -8,7 +8,7 @@ cd $ROOT_DIR
 pwd
 
 export DEPLOY_DB_VERSION=1
-export PACKAGE=https://releases.aeternity.io/aeternity-${DEPLOY_VERSION=latest}-ubuntu-x86_64.tar.gz
+export PACKAGE=https://releases.aeternity.io/aeternity-${DEPLOY_VERSION:-}-ubuntu-x86_64.tar.gz
```
venimus, line 11: is this intentional?

dincho: yes, it is in it's own commit. This default got me nothing but trouble during releases, as we deploy to our nodes before the release is actually declared latest, and of course I always forgot to specify concrete version.

venimus (approved):

venimus (approved):

IvayloPapazov (approved):

PR #558, opened 2022-03-01
Add docker buildx plugin

dincho:


mmpetarpeshev (commented):
👍

venimus (approved):

michalbagrowski (approved):

dincho:
@michalbagrowski our policy is that the PR authors should merge their PRs, that's for a reason that only the author knows the best moment for that.

In this case, I'm not sure we need this in general and also docker client version bump could cause incompatibility issues

PR #564, opened 2022-05-26
Don't restart SSH in CI

dincho:


dincho:
the handle condition doesn't seems to work 🤔

